PR #16, opened 2026-04-19
feat: track workflow startedAt and completedAt

Siddhant-K-code:
Records when a workflow first starts executing (`startedAt`) and when it reaches a terminal state (`completedAt`).

**Why:** there's no way to know how long a workflow took. This is the baseline needed for performance testing — especially useful when running a chaos matrix.

**What changed**
- `Workflow` interface gains `startedAt?` and `completedAt?` (ISO strings)
- Schema migration adds `started_at` / `completed_at` columns — additive, safe on existing DBs
- `syncWorkflowStatus` sets `startedAt` once on first running task, `completedAt` once on terminal state
- `serializeWorkflow` emits both fields plus a computed `durationMs`
- `formatWorkflowList` shows elapsed time next to each workflow (e.g. `2m 30s`)

8 tests added.

### DIFF
--- a/packages/app/src/__tests__/workflow-duration.test.ts
+++ b/packages/app/src/__tests__/workflow-duration.test.ts
@@ -1,0 +1,82 @@
+import { describe, it, expect } from 'vitest';
+import { serializeWorkflow, formatWorkflowList } from '../formatter.js';
+import type { Workflow } from '@invoker/data-store';
+
+function makeWorkflow(overrides: Partial<Workflow> = {}): Workflow {
+  return {
+    id: 'wf-1',
+    name: 'Test Workflow',
+    status: 'completed',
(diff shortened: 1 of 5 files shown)


EdbertChan:
Unsure about this. Because of the retry/recreate logic we have for workflows that recreate and retry tasks, these values can technically change.

We'd have to be clear that the "startedAt" and "completedAt" for the latest series of attempts.

would it not be better to derive this data from the tasks?

Siddhant-K-code:
> would it not be better to derive this data from the tasks?

yeah, that could be another way!

EdbertChan:
Not going to close this out yet. Legibility of execution is one of the promises of Invoker. Execution metadata would be indeed useful at some point. 

Let me come up with better architecture for this

PR #23, opened 2026-04-20
Top up scheduler after capacity-changing mutations

EdbertChan:
## Summary
Capacity-changing mutations like failed auto-fix, retries, and edits could leave runnable tasks sitting in `ready`/`pending` with idle capacity because the app changed workflow state but did not top the scheduler back up.

## Exact Failure / Symptom
Observed failure mode: no thrown exception; tasks stayed `ready`/`pending` and did not relaunch after a capacity-changing mutation.

## Exact Test Setup
- App regression: `packages/app/src/__tests__/bridge-orchestrator-executor.test.ts`
- Exact regression: `a failed fix path leaves ready pending work idle until executeGlobalTopup runs`
- The test frees capacity through a failed auto-fix path, leaves another root task ready, and proves that task stays idle until the global top-up path runs.

## Root Cause
App-layer mutation entrypoints changed runnable capacity but did not consistently call the top-up path afterward, so the orchestrator had new runnable work and free slots but no dispatch nudge.

## Why This Fix Is Right
This fix centralizes the scheduler refill at the app mutation boundary. Alternatives like polling or sprinkling more ad hoc dispatch calls would keep the behavior fragile and easy to miss in new mutation paths.

## Validation
- API, main, and headless entrypoints now top up through the same helper.
- The exact bridge regression above covers the idle-ready regression directly.

## Traceability
- Commit: `e9f74fc`
- Commit message: `Top up scheduler after capacity-changing mutations`


EdbertChan:
One follow-up note on this PR: the behavioral fix is correct, but there is still a real structural smell in how we apply post-mutation scheduler reconciliation across surfaces.

Right now, app/UI, headless, and API paths still duplicate variants of the same shape:
- run mutation
- local try/catch / error translation
- conditionally top up / reconcile scheduling

That duplication is the part most likely to drift again. I filed a follow-up to centralize it and also capture the related duplicated mutation-flow patterns we touched in this stack:
- INV-128 Centralize post-mutation scheduler reconciliation across app surfaces
- https://linear.app/invokerorchestrator/issue/INV-128/centralize-post-mutation-scheduler-reconciliation-across-app-surfaces

So: this PR keeps the behavioral fix narrow, and the refactor to remove the repeated mutation-wrapper smell is tracked separately in INV-128.

EdbertChan:
Another code-smell note on this PR: `packages/app/src/main.ts` is also carrying too many raw mutable runtime registries directly, for example:

```ts
const taskHandles = new Map<string, { handle: ExecutorHandle; executor: Executor }>();
const launchingTasks = new Set<string>();
const guiMutationHandlers = new Map<string, (...args: unknown[]) => Promise<unknown>>();
```

These each represent real state, but having `main.ts` own them as top-level collections is a sign that the entrypoint is becoming an ad hoc runtime state container instead of just wiring components together.

I filed a follow-up for that design debt separately:
- INV-129 Extract runtime state trackers out of app main entrypoint
- https://linear.app/invokerorchestrator/issue/INV-129/extract-runtime-state-trackers-out-of-app-main-entrypoint

This is related to INV-128 but distinct:
- INV-128 = duplicated post-mutation reconciliation logic across surfaces
- INV-129 = runtime state ownership / missing abstractions around execution tracking and GUI mutation routing

So this PR keeps the bug fix narrow, and the runtime state refactor is explicitly tracked in INV-129.

PR #24, opened 2026-04-20
Add SSH cleanup and merge lease repro coverage

EdbertChan:
## Summary
We had two real SSH managed-worktree proof gaps: stale `~/.invoker` cleanup misses and confusion about whether recreate was reusing the same branch/worktree identity.

## Exact Failure / Symptom
Exact failures covered by the new proofs:
- `... is already used by worktree at ...`
- `cannot force update the branch ... used by worktree at ...`

## Exact Test Setup
- SSH git-exec regression: `packages/execution-engine/src/__tests__/ssh-git-exec.test.ts`
  - `removes a stale worktree when canonicalRemoteWt starts with quoted tilde`
- SSH executor regression: `packages/execution-engine/src/__tests__/ssh-executor.test.ts`
  - `changes managed branch and worktree when request salt changes, as recreate does`

Together these pin down the two key facts: stale `~/.invoker` cleanup was real, and recreate itself changes the managed branch/worktree identity instead of intentionally reusing the same one.

## Root Cause
The repo had logic touching these areas, but no deterministic proof of the remote worktree cleanup miss or of recreate’s branch/worktree identity behavior. That made later fixes easy to misattribute.

## Why This Fix Is Right
This PR is intentionally proof-first. The right first step was to pin down the failing shapes before changing runtime behavior.

## Validation
- Adds focused SSH regressions without changing production behavior.
- The exact tests above capture both the cleanup miss and the recreate-identity proof directly in code.

## Traceability
- Commit: `d583847`
- Commit message: `Add SSH cleanup and merge lease repro coverage`


EdbertChan:
Clarifying scope on this PR: this one is intentionally proof-first, not the behavioral fix.

It captures two bug families as deterministic repro coverage:
- SSH stale managed-worktree cleanup / startup-recreate misses
- merge lease / merge-stall failure shapes

The actual runtime fixes for those families land later in the stack in narrower follow-up PRs:
- merge handoff / lease family: #25 and #29
- SSH managed-worktree reconciliation family: #28, #36, and #38

So the overlap is intentional:
1. prove the exact failure shape here
2. then fix the proven sub-families in later PRs

I wanted to call that out explicitly so this PR is reviewed as repro coverage / evidence, not as the place where the runtime behavior is fully corrected.

PR #30, opened 2026-04-20
Make orchestrator dispatch authoritative

EdbertChan:
## Summary
The same started task could be launched twice because the orchestrator dispatched it through `taskDispatcher`, and app-layer callers also manually called `executeTasks(started)` on the same result.

## Exact Failure / Symptom
Observed symptom: duplicate `launch started` events for the same task, leading to races like `uv_cwd` during provisioning.

## Concrete Scenario
A GUI/headless/API caller asks the orchestrator to drain runnable work. The orchestrator starts a task and dispatches it through `taskDispatcher`, then the caller also manually launches the same started task from the returned `started` list.

## Exact Test Setup
- App-layer test: `packages/app/src/__tests__/bridge-orchestrator-executor.test.ts`
- Exact regression: `dispatches each runnable task only once when taskDispatcher is wired`
- The test wires the bridge/orchestrator/executor path so the task becomes runnable once and proves only one launch happens.

## Root Cause
Dispatch ownership was split across two layers. `drainScheduler()` already dispatched started tasks, but GUI/headless/API/top-up code still tried to dispatch the same tasks manually.

## Why This Fix Is Right
Making the orchestrator authoritative removes the ownership overlap. Alternatives like deduping lower in the executor would paper over the race while leaving two launch authorities in place.

## Validation
- Bridge, headless, API, and execution-engine tests now assert single-dispatch behavior.
- The exact regression above covers the double-dispatch scenario directly.

## Traceability
- Commit: `26fbcad`
- Commit message: `Make orchestrator dispatch authoritative`


EdbertChan:
Follow-up architecture note:

The behavior fix in this PR is still the right short-term change, but the review discussion exposed a real API smell: `Orchestrator` now exposes runtime dispatch-ownership knowledge (`hasTaskDispatcher()` / equivalent checks) even though that concern really belongs above the orchestrator boundary.

In other words, this PR correctly fixes double-dispatch, but it does so with a pragmatic guard rather than the final API shape. The longer-term cleanup should make dispatch ownership explicit without callers needing to ask the orchestrator whether it has runtime wiring attached.

Filed follow-up: INV-130 `Remove dispatch-ownership signaling from Orchestrator API`
https://linear.app/invokerorchestrator/issue/INV-130/remove-dispatch-ownership-signaling-from-orchestrator-api

Planned direction in that follow-up:
- prefer a decision-only orchestrator that returns started tasks
- keep dispatch ownership in the app/runtime layer
- remove `hasTaskDispatcher()`-style integration checks from callers

So this PR remains the narrow bug fix, and INV-130 tracks the architectural cleanup.

EdbertChan:
One more note on the exact change we discussed:

```ts
if (newlyStarted.length > 0 && !orchestrator.hasTaskDispatcher?.()) {
  // manually dispatch
}
```

This line is correct for the bug fix, but it also shows the API smell pretty directly:
- `newlyStarted.length > 0` is asking about scheduler output
- `hasTaskDispatcher()` is asking about runtime wiring / dispatch ownership

Those are different concepts, and the fact that callers have to combine them is the part that should be cleaned up later.

Also, this is why a rename like `hasTasksToDispatch()` would actually be worse: it would blur task presence with runtime configuration. The real follow-up is not just naming, it is removing this runtime-dispatch-ownership question from the `Orchestrator` surface entirely.

I added that exact context to INV-130 as the follow-up design task:
https://linear.app/invokerorchestrator/issue/INV-130/remove-dispatch-ownership-signaling-from-orchestrator-api

PR #34, opened 2026-04-20
Fence auto-fix sessions per task

EdbertChan:
## Summary
This PR fixes a real queued auto-fix bug: one auto-fix intent could already be live for a task while sitting in the workflow mutation queue, and another failed-trigger path could try to schedule auto-fix again for that same task before the first intent actually executed.

The result was:
- more than one live auto-fix lifecycle for one task
- old queued auto-fix work surviving after the task had already moved on

## Exact Failure / Symptom
Representative bad behavior:
- a merge gate is `failed`
- auto-fix intent A is queued behind another running workflow mutation
- while A is still live-but-queued, a second scheduling pass tries to enqueue auto-fix again for the same task
- before A finally runs, the merge gate moves to `review_ready`
- the old queued auto-fix wakes up stale and tries to dispatch against the wrong state

## Exact Test Setup
- App-layer test: `packages/app/src/__tests__/persisted-workflow-mutation-coordinator.test.ts`
- Exact regression: `keeps one live auto-fix session per task and skips stale queued dispatch after review_ready`
- The test blocks the workflow mutation queue, enqueues one live auto-fix intent, attempts to schedule a second one for the same task, transitions the task to `review_ready`, then releases the queue and proves the queued fix becomes a no-op skip.

## Root Cause
Auto-fix scheduling was edge-triggered on failed-task state, but there was no task-scoped fence across queued and running auto-fix intents.

## Why This Fix Is Right
This fixes the bug at the two boundaries that matter:
- enqueue time: prevents a second live auto-fix session from being created for the same task
- dispatch time: prevents an older queued auto-fix from running after the task has already left `failed`

## Validation
- The exact regression above covers duplicate suppression and stale queued-dispatch skip.
- Coordinator tests assert both the enqueue-time fence and the dispatch-time revalidation.

## Follow-up
A separate follow-up is now tracked for a stronger invariant around retry/recreate canceling active auto-fix work:
- `INV-131` `Make retry preempt active auto-fix sessions like recreate`

## Traceability
- Commit: `5bbee06`
- Commit message: `Fence auto-fix sessions per task`


EdbertChan:
Follow-up note from review discussion:

PR #34 fixes a real bug: multiple auto-fix lifecycles could exist for the same task because one auto-fix intent could already be live in the workflow mutation queue while another failed-trigger path tried to schedule auto-fix again. The repro for this PR is the exact queued sequence in `persisted-workflow-mutation-coordinator.test.ts`, not a vague retry/recreate story.

That said, review also surfaced a related but stronger follow-up invariant we do **not** fully enforce yet:
- recreate hard-preempts running workflow mutation intents
- retry currently behaves as a queue fence, but not the same hard-preempting cancel path

So there is a follow-up task to make retry/recreate cancel active auto-fix sessions consistently:
- INV-131 `Make retry preempt active auto-fix sessions like recreate`
- https://linear.app/invokerorchestrator/issue/INV-131/make-retry-preempt-active-auto-fix-sessions-like-recreate

This PR should stay scoped to the bug it actually fixes:
- one live auto-fix session per task
- stale queued auto-fix becomes a no-op skip

The stronger retry/recreate preemption guarantee is tracked separately in INV-131.

PR #35, opened 2026-04-20
Make merge clone paths atomically unique

EdbertChan:
## Summary
Merge clone directories were named with `<label>-<Date.now()>`, so two same-label clone requests in the same millisecond could collide on the same destination path.

## Exact Failure / Symptom
`fatal: destination path '.../merge-clones/<label>-<timestamp>' already exists and is not an empty directory`

## Exact Test Setup
- Execution-engine test: `packages/execution-engine/src/__tests__/create-merge-worktree.test.ts`
- Exact regressions:
  - `same-millisecond createMergeWorktree calls with the same label get unique clone paths`
  - `same-millisecond sequential createMergeWorktree calls with the same label do not collide`
- The tests freeze `Date.now()` so two same-label requests share the same millisecond and prove clone-path allocation stays unique.

## Root Cause
`Date.now()` was being used as the uniqueness mechanism for merge clone paths. That is not atomic and is not collision-proof under concurrent same-label clone creation.

## Why This Fix Is Right
Atomic directory allocation fixes the race at the source. Alternatives like retrying after clone failure would keep the collision-prone naming scheme in place and only patch the symptom.

## Validation
- The exact regressions above prove same-label same-millisecond calls return different clone paths.
- Merge-clone allocation now uses an atomically unique directory with a readable label prefix.

## Traceability
- Commit: `0f270c1`
- Commit message: `Make merge clone paths atomically unique`


EdbertChan:
Follow-up note on temp directory retention:

This PR fixes the real bug by making merge clone paths atomically unique. That solves the collision problem, but it does **not** change retention behavior: if the app is killed, crashes, or the machine shuts down before normal cleanup runs, the merge clone directory can still remain on disk under `~/.invoker/merge-clones/...` after restart.

That is acceptable for this PR because the scope here is uniqueness, not lifecycle cleanup. But it is still worth re-evaluating whether we want stronger cleanup/scavenging for abandoned merge clone directories.

Filed follow-up: INV-132 `Re-evaluate merge clone temp directory retention and cleanup`
https://linear.app/invokerorchestrator/issue/INV-132/re-evaluate-merge-clone-temp-directory-retention-and-cleanup

PR #38, opened 2026-04-20
Retry SSH stale branch-owner cleanup from setup failure (follow-up to a0eaedc "Preserve stale SSH branch-owner cleanup on probe failure")

EdbertChan:
## Summary
The earlier stale-owner cleanup fix was real but incomplete: some stale branch-owner paths were not visible during preflight discovery and only appeared in the actual `setupTaskBranch(...)` Git error.

## Exact Failure / Symptom
`fatal: '<branch>' is already used by worktree at '<ownerPath>'` or `cannot force update the branch ... used by worktree at '<ownerPath>'` during setup.

## Exact Test Setup
- Execution-engine test: `packages/execution-engine/src/__tests__/ssh-executor.test.ts`
- Exact regression: `retries setup after parsing a stale branch-owner path from setupTaskBranch failure`
- The test forces preflight discovery to miss the stale owner path, then makes `setupTaskBranch(...)` fail with Git stderr naming the real owner path and proves SSH cleans that path up and retries once.

## Root Cause
Preflight reconciliation was still missing one surviving shape. The stale owner path was only recoverable from the stderr of the failed branch setup itself.

## Why This Fix Is Right
This is the right bounded fallback because it uses the authoritative Git error when preflight misses a case. Alternatives like more heuristic preflight probing or infinite retries would either stay incomplete or become unsafe.

## Validation
- The exact SSH executor regression above covers the setup-failure fallback path.
- The fallback cleans the canonical and parsed owner paths and retries setup once.

## Traceability
- Commit: `a45d7b6`
- Commit message: `Retry SSH stale branch-owner cleanup from setup failure`


EdbertChan:
Follow-up note: this fix still relies on a temporary invariant that a managed branch maps to a single managed worktree for reuse/cleanup. That assumption is acceptable for the current SSH reconciliation path, but it should go away once Invoker supports multiple managed worktrees on the same branch.

We should track a follow-up ticket to replace branch-ownership-based reconciliation with explicit managed-worktree identity, so discovery / reuse / cleanup can handle multiple live worktrees on the same branch without treating sibling worktrees as stale owners.

PR #63, opened 2026-04-21
Test admin-bypass Mergify queue config

EdbertChan:
## Summary
This PR migrates merge-mode changes out of app-layer special casing and into an orchestrator-owned retry-class invalidation path. `Orchestrator.editTaskMergeMode` now handles same-mode no-ops, cancel-first behavior for active merge nodes, persistence of the new workflow merge mode, and reset via `restartTask`. `setWorkflowMergeMode` becomes a thin delegate when a merge node exists.

The motivation is to eliminate stale merge-policy execution. Before this change, an in-flight merge node could continue running under the old mode after a flip. The tradeoff is a stricter invalidation path for active merge nodes, but that is the correct consequence of changing execution policy mid-flight.

```mermaid
flowchart LR
  subgraph Before
    A[setWorkflowMergeMode] --> B[persist mergeMode in app layer]
    B --> C[restart only in selected states]
    C --> D[active merge node could keep old policy]
  end
```

```mermaid
flowchart LR
  subgraph After
    A[setWorkflowMergeMode] --> B[orchestrator.editTaskMergeMode]
    B --> C{same mode?}
    C -- yes --> D[return no-op]
    C -- no --> E{merge node active?}
    E -- yes --> F[cancelTask]
    E -- no --> G[persist mergeMode]
    F --> G
    G --> H[restartTask]
  end
```

## Test plan
- [x] GitHub Actions for this PR are green.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -11,12 +11,29 @@ queue_rules:
       - -draft
       - "#approved-reviews-by >= 1"
 
+  - name: admin-bypass
+    merge_method: squash
+    queue_conditions:
+      - base=master
+      - -draft
+      - label=admin-bypass
+
 pull_request_rules:
   - name: autoqueue approved PRs to master
     conditions:
       - base=master
       - -draft
       - "#approved-reviews-by >= 1"
+      - -label=admin-bypass
     actions:
       queue:
         name: default
+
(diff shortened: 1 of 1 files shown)


EdbertChan:
@mergifyio queue admin-bypass

PR #64, opened 2026-04-21
Harden plan-to-invoker for policy-matrix documents

EdbertChan:
## Summary
This change moves `fixPrompt` and `fixContext` edits onto the centralized invalidation seam instead of leaving them as bespoke fix-session inputs. The architecture now runs these mutations through `CommandService.editTaskFixContext` and `Orchestrator.editTaskFixContext`, which persist the config change, apply the existing `fixContext` invalidation policy, and retry the task when needed. That gives fix-specific inputs a typed home in task config and makes them observable through the same policy machinery as other execution-spec mutations.

The motivation is consistency: changing fix prompt and fix context materially changes how the next fix attempt executes, so it should follow retry-class semantics rather than being synthesized ad hoc inside the conflict resolver. The tradeoff is a broader public mutation surface, but in return the behavior becomes explicit, testable, and aligned with the chart’s cancel-first and generation-bump rules.

```mermaid
flowchart LR
  A[Fix prompt/context edit] --> B[Conflict resolver builds prompt ad hoc]
  B --> C[No explicit invalidation routing]
```

```mermaid
flowchart LR
  A[Fix prompt/context edit] --> B[CommandService.editTaskFixContext]
  B --> C[Orchestrator.editTaskFixContext]
  C --> D[applyInvalidation retry-class policy]
  D --> E[Cancel active fix if needed]
  E --> F[Retry task with updated config]
```

## Test plan
- [x] GitHub Actions for this PR are green.


EdbertChan:
Landed directly on  as  after rebasing the branch content onto current . GitHub was still reporting the PR as , but the current PR diff applied cleanly on top of ; the mergeability state was stale/confused, not a real content conflict.

EdbertChan:
Landed directly on `master` as `2250a06b` after rebasing the branch content onto current `master`.

GitHub was still reporting the PR as `DIRTY`, but the current PR diff applied cleanly on top of `master`; the mergeability state was stale/confused, not a real content conflict.

PR #66, opened 2026-04-24
Fix SSH worktree identity drift on retries

EdbertChan:
## Summary
This change teaches the SSH auto-fix path to repair stale workspace metadata when the persisted local path no longer exists but the remote branch still has a managed worktree. The executor now asks the remote host which worktree currently owns the branch, repairs `task.execution.workspacePath` if needed, and then spawns the agent in the repaired location.

## Exact Failure / Symptom
SSH retry and auto-fix flows could fail even though the remote branch and managed worktree were still valid, because the persisted workspace path no longer matched the real remote location.

## Root Cause
The retry path treated persisted workspace metadata as authoritative even after remote worktree rename or recovery events. When that path went stale, the executor had no repair step and failed before it could reach the still-valid remote branch owner.

## Approach
On the stale-path error path, ask the remote host which managed worktree owns the branch, repair the persisted workspace path, and continue the fix run in the repaired location.

## Main Changes
- probe remote managed worktrees by branch owner on the SSH retry path
- repair `task.execution.workspacePath` when the stored path is stale
- continue auto-fix in the repaired workspace instead of failing fast

## Why This Fix Is Right
The source of truth for SSH-managed worktrees is the remote host, not stale local metadata. Repairing metadata from the live remote branch owner is safer than pretending the path must never drift.

This adds one remote lookup on the error path, but only when the persisted path is already invalid. That is a good trade because the normal path stays fast and the recovery path becomes robust.

## Validation

## Traceability
- Commit message: `Fix SSH worktree identity drift on retries`

### DIFF
--- a/packages/app/src/__tests__/app-layer-handoff-repro.test.ts
+++ b/packages/app/src/__tests__/app-layer-handoff-repro.test.ts
@@ -1,0 +1,191 @@
+import { describe, it, expect, beforeEach } from 'vitest';
+import { createTestHarness, type TestHarness } from '@invoker/test-kit';
+import type { PlanDefinition } from '@invoker/workflow-core';
+import { dispatchStartedTasksWithGlobalTopup } from '../global-topup.js';
+
+const LINEAR_PLAN: PlanDefinition = {
+  name: 'Linear Handoff Repro',
+  onFinish: 'merge',
+  mergeMode: 'automatic',
+  baseBranch: 'master',
+  featureBranch: 'plan/linear-handoff',
+  tasks: [
+    { id: 'A', description: 'Task A', command: 'echo a' },
+    { id: 'B', description: 'Task B', command: 'echo b', dependencies: ['A'] },
+  ],
+};
+
+const PARALLEL_PLAN: PlanDefinition = {
+  name: 'Parallel Handoff Repro',
+  onFinish: 'merge',
+  mergeMode: 'automatic',
+  baseBranch: 'master',
+  featureBranch: 'plan/parallel-handoff',
+  tasks: [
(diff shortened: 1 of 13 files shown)


EdbertChan:
Transferred from EdbertChan/Invoker#51 issue comment 4301230842
Original author: @EdbertChan
Original timestamp: 2026-04-23T02:02:29Z
Original URL: https://github.com/EdbertChan/Invoker/pull/51#issuecomment-4301230842

This looks mostly superseded by later PRs in the stack.

I checked the later branches, and the core SSH branch/worktree logic introduced here gets reworked again in:
- #52 `Mix attemptId into branch hash and gate workspace cleanup`
- #110 `step 4: persist attempt.branch before git worktree add`

In particular, these files change again later:
- `packages/execution-engine/src/ssh-executor.ts`
- `packages/execution-engine/src/task-runner.ts`
- `packages/execution-engine/src/__tests__/ssh-executor.test.ts`

Those later changes replace the earlier branch-hash/salt flow with lifecycle-tag/content-hash naming, change cleanup behavior, and add early branch persistence before `git worktree add`.

What does not appear to get reworked later in this stack:
- `packages/execution-engine/src/conflict-resolver.ts`
- `packages/execution-engine/src/ssh-git-exec.ts`
- `packages/execution-engine/src/__tests__/remote-fix-worktree-repair.test.ts`
- `packages/execution-engine/src/__tests__/ssh-worktree-metadata-repro.test.ts`

So if the review concern is about the main `ssh-executor` retry/worktree-identity path, I think that is largely superseded later in the stack. The remote fix-path repair pieces seem to remain relevant on their own.

PR #67, opened 2026-04-24
Mix attemptId into branch hash and gate workspace cleanup

EdbertChan:
## Summary
This change makes each task attempt content-addressable on `attemptId`, so a retry or recreate produces a brand-new branch and worktree path instead of reusing the prior attempt’s canonical location. The execution engine and SSH executor still keep workspace cleanup available behind `INVOKER_ENABLE_WORKSPACE_CLEANUP`, but cleanup is no longer required for correctness once the branch hash is unique per attempt.

## Problem
Retry and recreate flows could collide with stale branch and workspace state from prior attempts because they reused the same canonical branch identity.

## Root Cause
Branch naming mixed stable task identity with attempt identity too weakly. That made cleanup a correctness requirement instead of just a hygiene step, because a new attempt could land on the same branch path as an old attempt.

## Approach
Mix `attemptId` into the branch hash so each attempt gets a distinct branch and workspace identity, then gate cleanup behind an env flag because correctness no longer depends on branch deletion timing.

## Main Changes
- include `attemptId` in branch hash inputs
- give each attempt a distinct canonical branch and worktree path
- keep cleanup available behind `INVOKER_ENABLE_WORKSPACE_CLEANUP` rather than requiring it by default

## Why This Fix Is Right
The real bug was branch identity collision, not insufficient cleanup. Making branch identity unique per attempt removes the collision class directly instead of relying on timely deletion to preserve correctness.

The tradeoff is that stale refs and worktrees can linger longer when cleanup stays disabled. That is acceptable because the system remains correct without cleanup, and operators can still opt into cleanup for disk hygiene.

## Before
```mermaid
flowchart LR
  A[Task retry or recreate] --> B[Reuse prior canonical branch]
  B --> C[Collide with stale worktree state]
  C --> D[Cleanup required for correctness]
```

## After
```mermaid
flowchart LR
  A[Task attempt] --> B[Hash inputs include workflow, task, and attemptId]
  B --> C[Compute unique branch hash]
  C --> D[New canonical worktree path]
  D --> E[Skip cleanup by default]
  E --> F[Optional cleanup only for disk hygiene]
```

## Verification

## Traceability
- Commit message: `Mix attemptId into branch hash and gate workspace cleanup`

### DIFF
--- a/packages/app/src/__tests__/rebase-and-retry.test.ts
+++ b/packages/app/src/__tests__/rebase-and-retry.test.ts
@@ -84,13 +84,24 @@ function makeTaskState(overrides: {
 
 describe('rebase-and-retry: pool mirror cleanup before restart', { timeout: 120_000 }, () => {
   let tmpDir: string;
+  let prevCleanupEnv: string | undefined;
 
   beforeEach(() => {
+    // This suite specifically validates managed-branch cleanup behavior, which
+    // is gated behind INVOKER_ENABLE_WORKSPACE_CLEANUP. Force-enable for the
+    // duration of the suite; production default (off) is covered elsewhere.
+    prevCleanupEnv = process.env.INVOKER_ENABLE_WORKSPACE_CLEANUP;
+    process.env.INVOKER_ENABLE_WORKSPACE_CLEANUP = '1';
     tmpDir = createTempRepo();
   });
 
   afterEach(() => {
     rmSync(tmpDir, { recursive: true, force: true });
+    if (prevCleanupEnv === undefined) {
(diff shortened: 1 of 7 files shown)


EdbertChan:
Transferred from EdbertChan/Invoker#52 issue comment 4301241369
Original author: @EdbertChan
Original timestamp: 2026-04-23T02:04:20Z
Original URL: https://github.com/EdbertChan/Invoker/pull/52#issuecomment-4301241369

Follow-up tracked in Linear: INV-137 Remove transitional workspace cleanup disable flag\nhttps://linear.app/invokerorchestrator/issue/INV-137/remove-transitional-workspace-cleanup-disable-flag

PR #118, opened 2026-04-24
collision-free branches: acquire by content and validate leftovers

EdbertChan:
## Summary
This teaches acquisition to recognize cache-equivalent leftover worktrees, validate that they are still usable before reusing them, and treat rare cross-`actionId` hash collisions as trace telemetry instead of warnings.

## Problem
Leftover worktrees that were actually reusable still looked like hard conflicts, while obviously stale leftovers could slip into the reuse path too optimistically.

## Root Cause
Pool acquisition logic was still oriented around path conflict and branch conflict rather than the new content-plus-lifecycle naming model, and it did not explicitly validate leftover candidates before trying to reuse them.

## Approach
Parse experiment branches on acquisition, search for same-content leftovers, validate a leftover worktree before reuse/rename, and record cross-`actionId` hash collisions in execution traces instead of warning noisily.

## Why This Fix Is Right
The pool is the correct layer to reconcile leftover workspace state against the new naming semantics. Reuse now happens only after a lightweight validity scan, which keeps the optimization while making the reuse decision easier to trust.

The acquisition path becomes more sophisticated, but that complexity belongs at the pool boundary because it already owns worktree lifecycle, reuse, and stale-leftover cleanup.

## Before
```mermaid
flowchart LR
  A[acquire worktree] --> B[git worktree add]
  B --> C[fail on existing leaked path]
```

## After
```mermaid
flowchart LR
  A2[acquire worktree] --> B2[parse branch]
  B2 --> C2[find same-content leftover]
  C2 --> D2[validate leftover]
  D2 --> E2[rename to lifecycle or recreate]
  B2 --> F2[record cross-actionId hash collision in trace]
```

## Verification

## Traceability
- Commit message: `collision-free branches: acquire-by-content + collision warn (step 3)`

### DIFF
--- a/packages/app/e2e/edit-task-command.spec.ts
+++ b/packages/app/e2e/edit-task-command.spec.ts
@@ -7,9 +7,24 @@
  */
 
 import { test, expect, loadPlan, startPlan, waitForTaskStatus, E2E_REPO_URL } from './fixtures/electron-app.js';
+import type { Page } from '@playwright/test';
 
 function findTaskByIdSuffix(tasks: Array<any>, taskId: string) {
   return tasks.find((t) => t.id === taskId || t.id.endsWith(`/${taskId}`));
+}
+
+async function selectTaskAndWaitForCommandEditor(page: Page, taskSuffix: string): Promise<void> {
+  const node = page.locator(`.react-flow__node[data-testid$="/${taskSuffix}"]`);
+  const commandDisplay = page.locator('[data-testid="command-display"]');
+  for (let attempt = 0; attempt < 3; attempt++) {
+    await node.click();
+    try {
+      await expect(commandDisplay).toBeVisible({ timeout: 3000 });
+      return;
+    } catch (err) {
+      if (attempt === 2) throw err;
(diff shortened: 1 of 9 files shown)


EdbertChan:
Transferred from EdbertChan/Invoker#109 review comment 3128631989
Original author: @EdbertChan
Original timestamp: 2026-04-23T05:56:25Z
Original URL: https://github.com/EdbertChan/Invoker/pull/109#discussion_r3128631989

Context: `packages/app/src/__tests__/bridge-orchestrator-executor.test.ts` line 937 (RIGHT)

Moved from source comment: Steps 11 → 14 (`docs/architecture/task-invalidation-roadmap.md`): `replaceTask` on a *live* workflow now routes through `forkWorkflow` (Step 14) instead of throwing `TopologyForkRequired` (Step 11). This test exercises the **in-place** replacement path on purpose, so it cancels the live downstream subgraph (B drags C with it) to make the workflow terminal first. The assertions about the in-place replacement subgraph below are unchanged.

PR #146, opened 2026-04-28
Task delta revision step 2: main-process gap detection and authoritative recovery

EdbertChan:
## Summary

This PR replaces the Electron main-process fallback path for task deltas with
explicit revision continuity checks and authoritative reloads from persistence.

The DB remains the source of truth, but the main process still keeps an
in-memory task cache because the UI delta stream is the hot path: applying
contiguous updates from memory is much cheaper than reloading every task from
SQLite on every event. The cache exists to serve low-latency incremental UI
updates, not to define correctness.

On top of step 1's revisioned contracts, the main process now treats
`lastKnownTaskStates` as a synchronization cache instead of a source of truth.
Ordered deltas still apply fast, but cache misses and revision gaps now
quarantine the task, reload the authoritative record through IPC, and resume
from the recovered revision boundary.

### Before

```mermaid
graph TD
    A[Main process receives task delta] --> B[Best-effort merge into cached task]
    B --> C[Unknown or out-of-order state handled heuristically]
    C --> D[Renderer may continue from stale continuity]
```

### After

```mermaid
graph TD
    A[Main process receives task delta] --> B[Check revision continuity against cache]
    B -->|contiguous| C[Apply delta normally]
    B -->|gap or cache miss| D[Quarantine task]
    D --> E[Load authoritative task by id via IPC]
    E --> F[Publish recovered task from persisted source of truth]
```

## Test Plan

- [ ] `pnpm --filter @invoker/app test` not run here (`node_modules` missing)
- [ ] `pnpm run test:all`

## Revert Plan

- Safe to revert? Yes
- Revert command: `git revert <merge_commit_sha>`
- Post-revert steps: None
- Data migration? No

### DIFF
--- a/packages/app/src/__tests__/delta-merge.test.ts
+++ b/packages/app/src/__tests__/delta-merge.test.ts
@@ -1,13 +1,17 @@
 /**
- * Tests for the delta-merge helper extracted from main.ts.
+ * Tests for the revision-aware delta-merge cache.
  *
- * Covers the out-of-order delta scenario: an `updated` delta arrives
- * for a task that has no prior `created` entry in the state map.
- * The fix fetches the task from the orchestrator fallback so the
- * update is not silently dropped.
+ * Covers:
+ * - created / removed deltas
+ * - updated deltas with correct revision continuity
+ * - revision gap → quarantine
+ * - unknown task → quarantine
+ * - quarantined task ignores further deltas until resolved
+ * - resolveQuarantine re-seeds the cache from authoritative data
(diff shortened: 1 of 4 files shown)


EdbertChan:
Known limitation: step 2 intentionally accepts the per-task authoritative reload cost on quarantine in exchange for correctness on cache gaps. If continuity breaks broadly, many tasks could quarantine and drive repeated `loadTask(taskId)` reads. We are punting any batching/coalescing/rate-limiting strategy for widespread quarantine to follow-up work so this stack can land the correctness model first.

PR #183, opened 2026-05-04
Retry reset semantics should match recreate immediacy

EdbertChan:
## Summary
Change Invoker retry semantics so retry behaves like recreate in reset scope and timing, while still remaining a retry-class action.
Today workflow retry is explicitly incremental: it preserves completed tasks outside invalidated failed-root subgraphs, which is why bulk retry does not make an entire workflow immediately pending again. The desired behavior is that retry instantly resets the relevant scope to `pending`, just like recreate, so operators can rely on immediate visible reset semantics from the UI, headless retry, and bulk retry scripts.

Preserve retry lineage where it matters: keep branch and workspace metadata instead of doing recreate's full lineage wipe, and do not upgrade retry into fresh-base rebase semantics. The implementation must also ensure queued or in-flight retry-scope mutations cannot keep stale work alive after the retry reset lands.

This repository is Invoker itself, so keep `onFinish: pull_request` with `mergeMode: external_review`; after implementation commits are ready, publish/update the resulting stack with `mergify stack push`.

---
Retry reset semantics should match recreate immediacy — 6 tasks completed, 0 failed, 0 skipped

<details>
<summary>Task breakdown</summary>

| Task | Description | Status |
|------|-------------|--------|
| wf-1777324024407-4/add-retry-regression-coverage | Add deterministic regression coverage for recreate-style retry reset semantics across workflow-core and app surfaces. | completed |
| wf-1777324024407-4/implement-task-retry-reset | Change task-scope retry to reset immediately like recreate while preserving retry lineage metadata. | completed |
| wf-1777324024407-4/implement-workflow-retry-reset | Change workflow-scope retry to reset the entire workflow immediately like recreate while preserving retry lineage metadata. | completed |
| wf-1777324024407-4/run-workflow-core-tests | Run the workflow-core package test suite after changing retry reset semantics. | completed (passed) |
| wf-1777324024407-4/run-app-retry-tests | Run targeted app retry and headless tests that verify retry command-surface and dispatch behavior. | completed (passed) |
| wf-1777324024407-4/rerun-postfix-retry-surface | Re-run retry-focused workflow-core tests as final post-fix verification of recreate-style reset behavior. | completed (passed) |

</details>

<details>
<summary>File changes per task</summary>

### wf-1777324024407-4/add-retry-regression-coverage — Add deterministic regression coverage for recreate-style retry reset semantics across workflow-core and app surfaces.
.../__tests__/bridge-orchestrator-executor.test.ts |  37 +++++-
[...]

PR #184, opened 2026-05-04
Set up ESLint with no-process-exit rule

EdbertChan:
## Summary

This PR adds a repo-wide ESLint baseline with a narrow initial goal: surface unsafe `process.exit()` usage and make lint part of the existing verification pipeline. It introduces a flat `eslint.config.mjs`, installs the new lint dependencies at the root, adds a dedicated `lint` script, and wires lint into `check:all` so the new rule runs anywhere the standard repo checks run.

To keep the rule actionable, the PR annotates the intentional `process.exit()` calls in the Electron entrypoint and interactive sandbox instead of weakening the rule globally. The review surface also includes a small remote-execution hardening change in `ssh-executor.ts`: commands sent to remote hosts now strip a leading local `cd /absolute/path && ...` prefix before execution so author-machine paths do not leak into SSH runs.

### Before

```mermaid
graph TD
    DEV[Developer or CI]
    CHECK[check:all]
    DEPS[check:deps]
    TYPES[check:types]
    OWNER[check-owner-boundary]
    REVIEW[Manual review of process.exit call sites]
    SSH[SSH executor forwards authored command verbatim]
    REMOTE[Remote host may receive cd /local/path && ...]

    DEV --> CHECK
    CHECK --> DEPS
    CHECK --> TYPES
    CHECK --> OWNER
    CHECK -. no lint stage .-> REVIEW
    SSH --> REMOTE

    style REVIEW fill:#ffcdd2
    style REMOTE fill:#ffcdd2
```

### After

```mermaid
graph TD
    DEV[Developer or CI]
    CHECK[check:all]
    DEPS[check:deps]
    TYPES[check:types]
    OWNER[check:owner-boundary]
    LINT[pnpm lint]
    CFG[eslint.config.mjs<br/>TS recommended rules<br/>+ no-process-exit warn]
    EXC[Intentional exits annotated inline]
    SSH[SSH executor sanitizes command prefix]
    REMOTE[Remote host runs command from managed worktree cwd]

    DEV --> CHECK
    CHECK --> DEPS
    CHECK --> TYPES
    CHECK --> OWNER
    CHECK --> LINT
    LINT --> CFG
    CFG --> EXC
    SSH --> REMOTE

    style LINT fill:#e8f5e9
    style CFG fill:#e3f2fd
    style REMOTE fill:#e8f5e9
```

## Test Plan

- [x] `pnpm lint`
- [x] `pnpm check:all`

## Revert Plan

- Safe to revert? Yes
[...]

PR #191, opened 2026-05-04
Reduce DAG visual effects and add perf guard tests

EdbertChan:
## Summary

Reduces expensive visual effects in the workflow DAG during viewport interaction and adds a guard test that locks in the cheaper styling choices.

The graph branch removes heavy node shadows, edge drop shadows, and subtree-wide animation or filter effects from the Electron DAG surface, while keeping the existing graph behavior and adding a source-level test that prevents those expensive styles from being reintroduced accidentally.

## Test Plan

- [x] `cd packages/ui && pnpm exec vitest run src/__tests__/task-dag-perf-guard.test.ts`
- [x] `bash scripts/test-suites/optional/41-electron-ui-perf.sh`

## Revert Plan

- Safe to revert? Yes
- Revert command: `git revert 862bfb22`
- Post-revert steps: None
- Data migration? No

Depends-On: #183

### DIFF
--- a/packages/ui/src/__tests__/task-dag-perf-guard.test.ts
+++ b/packages/ui/src/__tests__/task-dag-perf-guard.test.ts
@@ -1,0 +1,47 @@
+import { describe, expect, it } from 'vitest';
+import { readFileSync } from 'node:fs';
+import { resolve } from 'node:path';
+
+const taskNodeSource = readFileSync(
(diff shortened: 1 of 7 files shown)


EdbertChan:
Closing to mirror the current fork review state for origin PR #184. We will reopen later if needed.

PR #447, opened 2026-05-12
Fix fixing-with-AI Step 4: make provisioning engine failures legible

EdbertChan:
## Summary

When worktree provisioning fails because the machine is on an unsupported Node version, the current error is too noisy to act on quickly.

This PR makes that failure path human-readable by detecting `ERR_PNPM_UNSUPPORTED_ENGINE` and surfacing:
- expected Node version
- current Node version
- direct machine fix guidance (`nvm install 22 && nvm use 22`)

The full provisioning output is still included (trimmed) so deep debugging remains possible.

## Test Plan

- [x] `PATH="/opt/homebrew/opt/node@22/bin:$PATH" npm exec pnpm -- test -- "src/__tests__/worktree-executor.test.ts" -t "fails provisioning when timeout elapses and kills the process group"` (targeted run; package suite still includes existing unrelated baseline failures)
- [x] `PATH="/opt/homebrew/opt/node@22/bin:$PATH" bash scripts/repro/repro-launching-stall.sh`

## Revert Plan

- Safe to revert? Yes.
- Revert command: `git revert f0fd7b5`
- Post-revert steps: None.
- Data migration? No.

### DIFF
--- a/packages/execution-engine/src/worktree-executor.ts
+++ b/packages/execution-engine/src/worktree-executor.ts
@@ -46,6 +46,7 @@ export interface WorktreeExecutorConfig {
 }
 
 const DEFAULT_WORKTREE_PROVISION_TIMEOUT_MS = 15 * 60 * 1000;
+const MAX_PROVISION_OUTPUT_CHARS = 6_000;
 
 function resolveWorktreeProvisionTimeoutMs(): number {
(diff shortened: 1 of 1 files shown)


EdbertChan:
Closing as won't do. We should not classify provisioning failures or add Node-specific interpretation here; provisioning can fail for many reasons, and the right behavior is to keep the raw provisioning output legible and intact as-is.

PR #839, opened 2026-05-19
CodeBuddy: enforce SSH pool concurrency limit

EdbertChan:
## Problem

Two compounding bugs allowed unlimited concurrent tasks to land on a single SSH pool member even when `maxConcurrentTasksPerMember` was configured, causing the SSH host overload seen after `rebase-recreate-all`.

**Root cause in production:** After dispatching 37 workflows simultaneously on a single-member SSH pool, in-memory load maps were empty (fresh process state), so every task saw `load=0` on the same host. 9 concurrent SSH sessions landed on `remote_digital_ocean_1` (157.230.133.215), causing .git/config lock contention, heartbeat stalls, and cascading autofix triggers.

---

## Bug 1 — over-limit fallback in `selectPoolMember()`

**File:** `packages/execution-engine/src/task-runner.ts`

When `leastLoaded` strategy was active and all members were at capacity, the fallback returned the full `scored` array instead of `undefined`, making the limit advisory rather than enforced.

```typescript
// BEFORE (broken): returns over-limit members when all are full
const candidates = scored.some((e) => e.hasCapacity)
  ? scored.filter((e) => e.hasCapacity)
  : scored;

// AFTER: hard block
const candidates = scored.filter((e) => e.hasCapacity);
if (candidates.length === 0) return undefined;
```

---

## Bug 2 — `Promise.all` race in `executeTaskInner()`

`executeTasks()` dispatches all ready tasks concurrently via `Promise.all`. SSH `executor.start()` takes 30+ seconds to complete. Tasks 2–N all enter `selectPoolMember()` before task 1 has finished registering its `pendingPoolSelections` entry, so they all observe `load=0` for the same member.

**Fix:** Introduce a per-pool promise-chain mutex (`poolSelectionLocks: Map<string, Promise<void>>`) via a new `withPoolSelectionLock(poolId, fn)` method. The `selectExecutor()` call — which atomically queries `selectPoolMember()` and writes `pendingPoolSelections` — is now serialised per `poolId` before `executor.start()` is awaited.

```typescript
executor = await this.withPoolSelectionLock(
  task.config.poolId ?? `_nopool_${task.id}`,
  () => this.selectExecutor(task, attemptedPoolMemberKeys),
);
```

---

## Tests

Two new `describe` blocks in `task-runner.test.ts`:

- **`selectPoolMember – maxConcurrentTasksPerMember hard cap (Bug 1)`** — verifies `undefined` is returned when a single-member pool is saturated; verifies the least-loaded member is selected when one has capacity.
- **`withPoolSelectionLock – serializes concurrent selections (Bug 2)`** — verifies that two concurrent `Promise.all` selections on a single-member pool result in exactly one successful selection and one `undefined`, not two successes.

**Result:** 974/974 tests pass, `tsc` clean.

### DIFF
--- a/packages/execution-engine/src/__tests__/task-runner.test.ts
+++ b/packages/execution-engine/src/__tests__/task-runner.test.ts
@@ -5988,4 +5988,114 @@ console.log(JSON.stringify(out));
     });
   });
 
+  // -------------------------------------------------------------------------
+  // SSH pool concurrency-limit regression tests
+  // Bug 1: selectPoolMember() over-limit fallback returned over-limit members
+  //        instead of undefined, making maxConcurrentTasksPerMember advisory.
+  // Bug 2: Promise.all dispatch raced before pendingPoolSelections was updated,
+  //        so all concurrent tasks observed load=0 for the same pool member.
+  // -------------------------------------------------------------------------
+
+  describe('selectPoolMember – maxConcurrentTasksPerMember hard cap (Bug 1)', () => {
+    it('returns undefined when all members are at the concurrency limit', () => {
+      const tasks = new Map<string, TaskState>();
+      const runner = createExecutorWithTasks(tasks);
+
+      const pool = {
+        members: [{ type: 'ssh' as const, id: 'host-1', host: '1.2.3.4', user: 'root' }],
+        selectionStrategy: 'leastLoaded' as const,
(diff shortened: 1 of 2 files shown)


EdbertChan:
Closing as superseded. SSH pool capacity is already enforced on current master.

PR #1021, opened 2026-06-02
Fix intent cancellation Step 6: worker response lineage guard (regression gated v2)

EdbertChan:
Validation passed. Here is the final PR body:

## Summary

Worker responses that carried a matching `attemptId` were accepted without checking `executionGeneration`. A stale generation of the same attempt could therefore mutate the live task graph.

This closes that gap. `handleWorkerResponse` now rejects any response whose `executionGeneration` is present but does not match the live task's active generation, even when the `attemptId` matches.

Older producers that omit `executionGeneration` stay accepted, so backwards compatibility holds. When both fields are present, both must match for the response to apply.

Rejected stale responses are logged as `STALE_GENERATION_REJECTED`, beside the existing `STALE_ATTEMPT_REJECTED` and `SUPERSEDED_ATTEMPT_REJECTED` guards.

This is Step 6 of the intent-cancellation stack and follows the Step 5 review gate (wf-1778583486424-15).

### Before

```mermaid
graph TD
    A[Worker response] --> B{attemptId matches selectedAttemptId?}
    B -- no --> R[Reject: STALE_ATTEMPT]
    B -- yes --> C{attempt discarded?}
    C -- yes --> S[Reject: SUPERSEDED_ATTEMPT]
    C -- no --> APPLY[Apply to task graph]
```

### After

```mermaid
graph TD
    A[Worker response] --> B{attemptId matches selectedAttemptId?}
    B -- no --> R[Reject: STALE_ATTEMPT]
    B -- yes --> C{attempt discarded?}
    C -- yes --> S[Reject: SUPERSEDED_ATTEMPT]
    C -- no --> G{executionGeneration present and not active?}
    G -- yes --> T[Reject: STALE_GENERATION]
    G -- no --> APPLY[Apply to task graph]
```

## Test Plan

- [ ] `pnpm --filter @invoker/workflow-core exec vitest run src/__tests__/orchestrator.test.ts src/__tests__/orchestrator-dispatcher.test.ts`
- [ ] `pnpm run test:all`

## Revert Plan

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

### DIFF
--- a/packages/workflow-core/src/__tests__/orchestrator.test.ts
+++ b/packages/workflow-core/src/__tests__/orchestrator.test.ts
@@ -7088,6 +7088,55 @@ describe('Orchestrator', () => {
       expect(orchestrator.getTask(taskId)?.status).toBe('running');
     });
 
+    it('rejects a response whose attempt id matches but whose execution generation is stale', () => {
+      orchestrator.loadPlan({
+        name: 'attempt-match-stale-generation-rejection',
+        tasks: [{ id: 't1', description: 'Task 1' }],
+      });
+      orchestrator.startExecution();
+
+      const taskId = sid(orchestrator, 0, 't1');
+      const runningTask = orchestrator.getTask(taskId)!;
+      const activeAttemptId = runningTask.execution.selectedAttemptId!;
+      const staleGeneration = runningTask.execution.generation ?? 0;
+
+      // Establish known live lineage, then advance ONLY the execution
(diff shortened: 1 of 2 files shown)


EdbertChan:
Superseded by a resubmitted Invoker workflow chain starting at Step 2, with Step 2 gated on Step 1 and each later step gated on the previous step.

PR #1090, opened 2026-06-05
[CONTROVERSIAL] UI Perf Issue 3: Reduce large SQLite flush stalls

EdbertChan:
## Summary

Reduce startup-adjacent SQLite flush stalls by moving file-backed task output payloads out of the sql.js database export path. The adapter now stores output spool bytes in a sidecar log per task while keeping offsets, byte lengths, and storage metadata in SQLite, so small task updates no longer re-export hundreds of MiB of output text.

Add a deterministic large-flush repro that seeds a temporary database, performs a measured write, and enforces the optimized thresholds for exported bytes and flush time. Preserve legacy output reads, snapshot sidecar output with database backups, and remove the duplicate headless output persistence path so task output flows through the unified output API.

### Before

```mermaid
graph TD
    Runner[Task runner output] --> Headless[Headless/app callbacks]
    Headless --> TaskOutput[task_output rows]
    Headless --> OutputSpool[output_spool.data rows]
    OutputSpool --> SqlJs[sql.js in-memory database]
    TaskUpdate[Small task status update] --> SqlJs
    SqlJs --> Export[Full database export on flush]
```

### After

```mermaid
graph TD
    Runner[Task runner output] --> Unified[appendTaskOutput / appendOutputChunk]
    Unified --> Sidecar[per-task output sidecar log]
    Unified --> Metadata[output_spool offset, byte_length, storage]
    Metadata --> SqlJs[sql.js in-memory database]
    TaskUpdate[Small task status update] --> SqlJs
    SqlJs --> Export[Compact database export on flush]
    Snapshot[DB snapshots] --> SidecarCopy[copy output-spool sidecars]
```

## Test Plan

- [x] `bash scripts/repro/repro-sqlite-large-flush.sh && pnpm --filter @invoker/data-store test && pnpm --filter @invoker/app exec playwright test e2e/startup-nonempty-responsiveness.spec.ts`
- [x] `pnpm run test:all`

## Revert Plan

- Safe to revert? Yes
- Revert command: `git revert -m 1 81045361`
- Post-revert steps: Run `pnpm run test:all` and confirm task output remains readable for any databases written while this PR was active.
- Data migration? No manual migration; the adapter adds nullable/default `output_spool` columns on open and sidecar files are runtime persistence artifacts.

### DIFF
--- a/packages/app/e2e/global-setup.ts
+++ b/packages/app/e2e/global-setup.ts
@@ -5,7 +5,7 @@
  * so WorktreeExecutor can clone without a network. Sharded CI can override the
  * bare-repo path via INVOKER_E2E_BARE_REPO to avoid cross-shard interference.
  */
-import { execSync } from 'child_process';
+import { execSync, spawn, type ChildProcess } from 'child_process';
 import { existsSync, rmSync } from 'fs';
 import * as path from 'path';
 import { resolveRepoRoot } from '@invoker/contracts';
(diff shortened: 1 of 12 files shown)


EdbertChan:
This is controversial because we are effectively re-introducing caching: output bytes now live in sidecar persistence while SQLite keeps the offsets/length/storage metadata. That avoids the sql.js full-export stall, but it also adds cache-like invalidation and consistency concerns around backup/restore and task output reads.\n\nA cleaner long-term direction is to migrate persistence to native SQLite, which should let us keep task output in the database without paying the sql.js whole-database export cost. If we do that migration, we can likely remove this sidecar output path instead of maintaining another persistence layer.

EdbertChan:
Follow-up tracking ticket: [INV-186](https://linear.app/invokerorchestrator/issue/INV-186/migrate-persistence-to-native-sqlite-and-remove-sidecar-output-caching).

PR #1293, opened 2026-06-11
fix(data-store): clearing a workflow's external dependencies was silently ignored

EdbertChan:
## Summary

When a workflow is deleted, Invoker removes it from each dependent's "waiting on" list.

The database layer ignored that removal: "set the list to nothing" looked identical to "no change requested." So dependents waited forever on a deleted workflow (live example: WF-INV-143).

Fix: `updateWorkflow` now treats `externalDependencies: undefined` as "clear the list." Omitting the key still means "leave it unchanged." Same for `externalDependencyChanges`.

## Test Plan

- [x] `cd packages/data-store && pnpm exec vitest run src/__tests__/delete-workflow-dependent-cleanup.test.ts` (fails before the fix, passes after)
- [x] `cd packages/app && pnpm test`
- [x] `pnpm run check:types`

## Revert Plan

- Safe to revert? Yes
- Revert command: `git revert 85628580e`
- Post-revert steps: None — reverting restores the dangling-dependency bug; rows already cleared stay cleared (no schema change).
- Data migration? No. Past deletions are not auto-repaired; run `--headless detach-workflow <dependent> <deleted-upstream>` per affected workflow after deploying.

### DIFF
--- a/packages/data-store/src/__tests__/delete-workflow-dependent-cleanup.test.ts
+++ b/packages/data-store/src/__tests__/delete-workflow-dependent-cleanup.test.ts
@@ -1,0 +1,107 @@
+/**
+ * Repro: deleting (or detaching from) an upstream workflow must clear the
+ * dependent's externalDependencies in the real SQLite store.
+ *
+ * `Orchestrator.detachWorkflowInternal` clears the last dependency by passing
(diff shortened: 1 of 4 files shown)


EdbertChan:
@mergifyio queue

EdbertChan:
@Mergifyio queue admin-bypass

PR #1304, opened 2026-06-13
[State Invariants](1) Add task reset rulebook

EdbertChan:
## Summary

Task resets now use one shared rulebook instead of copied field sets.

The old copies could drift as task state grew. This makes each reset caller ask the same owner what to clear.

<details>
<summary>Review metadata</summary>

Review Claim: The reset rulebook correctly replaces duplicated task reset field sets.

Review Lane: refactor

Review Unit: ownership-refactor

Safety Invariant: Each caller still selects the same reset kind and writes the same task shape.

Slice Rationale: This is the base slice because later checks depend on one reset owner.

</details>

## Non-goals

- Does not change scheduling behavior.
- Does not change stored task shape.
- Does not change workflow invalidation planning.

### Before

```mermaid
graph TD
    A["Reset caller"] --> B["Local copied field set"]
    B --> C["TaskStateChanges"]
```

### After

```mermaid
graph TD
    A["Reset caller"] --> B["buildTaskResetChanges(kind)"]
    B --> C["Shared reset rulebook"]
    C --> D["TaskStateChanges"]
```

## Test Plan

- [x] `pnpm --filter @invoker/workflow-core test -- task-reset-policy.test.ts`
- [x] `pnpm --filter @invoker/workflow-core build`
- [x] `pnpm run check:types`
- [x] `bash scripts/test-suites/required/26-reset-rulebook-proof.sh`

## Revert Plan

- Safe to revert? Yes
- Revert command: `git revert 1622e8bfc`
- Post-revert steps: None.
- Data migration? No.

PR #1364, opened 2026-06-14
ci: consolidate e2e into one sharded, blocking proof gate

EdbertChan:
## Summary

Collapses the e2e CI surface into a single sharded, blocking gate. Today the same e2e-on-mock-DB suites run twice: in the blocking `dry-run / case-*` jobs **and** in the 17-min, non-blocking `inv117 / deterministic proof`, which globs all of `required/` (vitest, static checks, e2e — a dumping ground) and doesn't even gate merges.

This replaces both with **`e2e-proof`** (4 shards) + a blocking **`e2e-proof / aggregate`** gate, driven by an explicit `proof-e2e.manifest` of the 7 true e2e-on-mock-DB suites. Vitest/static/UI keep their own dedicated jobs. The aggregate validates the full proof contract (exact inventory + zero failures) over the merged per-shard state, so coverage is preserved while wall-clock drops from ~17 min serial to the slowest shard.

## Notes

- `run-all-tests.sh` proof mode now reads the manifest instead of globbing `required/`, derives its inventory count from the manifest (no more hardcoded counts), and supports shard slicing + an aggregate pass.
- `.mergify.yml`: the `dry-run / case-*` merge gates are replaced by `e2e-proof / aggregate`. The `master` ruleset pins no status checks, so the swap is safe.
- Base of the state-invariants stack — #1304 depends on this.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,13 +102,17 @@ jobs:
       - name: Run quality check
         run: ${{ matrix.command }}
 
-  inv117-proof:
+  e2e-proof:
     needs: build-artifacts
     runs-on: ubuntu-latest
     container:
       image: mcr.microsoft.com/playwright:v1.58.2-noble
-    timeout-minutes: 120
-    name: inv117 / deterministic proof
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [0, 1, 2, 3]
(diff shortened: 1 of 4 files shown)


EdbertChan:
Closed in favor of #2931, rebuilt from current master with the current case-2a/case-2b e2e split.

PR #1384, opened 2026-06-15
Fix intent cancellation Step 2: fix flow lineage guards (restacked v3)

EdbertChan:
## Summary

This makes the fix-with-agent and conflict-resolution flows throw away their results when the task has already moved on.

Before, a slow fix could finish after the user retried or recreated a task. The late result would then overwrite newer task output, fix approval, branch metadata, or conflict cleanup.

That happened because killing the underlying process is best-effort. Remote or agent work can still return after the task was invalidated.

The fix records the task's identity, its selected attempt and generation, when the flow starts. It re-checks that identity, and the abort signal, before every write.

If the task moved to a newer attempt or generation, or was aborted, the flow stops with a stale-lineage error and writes nothing.

This is part 2 of the intent-cancellation stack, restacked.

### Before

```mermaid
graph TD
    A[Start fix flow] --> B[Run async fix or resolve]
    B --> C[Write output, approval, branch, cleanup]
    C --> D[Task state overwritten even if a newer attempt exists]
```

### After

```mermaid
graph TD
    A[Start fix flow] --> S[Capture lineage: attempt and generation]
    S --> B[Run async fix or resolve]
    B --> G{Lineage current and not aborted?}
    G -- yes --> C[Write output, approval, branch, cleanup]
    G -- no --> X[Throw StaleLineageError, no writes]
```

## Test Plan

- [ ] `pnpm run test:all`
- [ ] `cd packages/app && pnpm test -- workflow-actions`

## Revert Plan

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No


EdbertChan:

PR #1390, opened 2026-06-16
Fix intent cancellation Step 3: SSH startup lineage guards (restacked v3)

EdbertChan:
Validation passes with no warnings. Final PR body:

## Summary

This adds a regression test that locks in a fix for stale SSH task launches.

When a task is retried, a slow first attempt could still fail later. That late failure used to overwrite the live task with old workspace and branch info.

The guard that stops this already shipped in earlier steps of this stack. This step proves it works and keeps it from breaking again.

The new test launches attempt one, moves the task to attempt two, then lets attempt one fail. It checks the stale path, branch, and failure never reach the live task.

## Test Plan

- [ ] `pnpm run test:all`

## Revert Plan

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

### DIFF
--- a/packages/execution-engine/src/__tests__/ssh-worktree-metadata-repro.test.ts
+++ b/packages/execution-engine/src/__tests__/ssh-worktree-metadata-repro.test.ts
@@ -144,4 +144,81 @@ describe('SSH worktree metadata repro', () => {
       }),
     }));
   });
+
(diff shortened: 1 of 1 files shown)


EdbertChan:
@Mergifyio refresh

PR #1398, opened 2026-06-16
Fix intent cancellation Step 4: durable failure diagnostics (regression gated v2)

EdbertChan:
## Summary

Startup failures now keep concrete stderr and stdout details in durable task output instead of collapsing later reads to a generic terminal error.

Forced-stop handling now records the same kind of diagnostics for every task state that can be shut down.

The test-only no-op provision hook keeps the regression suite bounded while covering the durable-output path.

<details>

Review Claim: Approve routing startup and forced-stop failures into durable task output as concrete diagnostic blocks.

Review Lane: behavior

Review Unit: routing

Safety Invariant: Task status and exit semantics stay unchanged. The slice only adds durable diagnostic text and a bounded test-only provisioning hook.

Slice Rationale: This slice isolates the durable-output wiring and tests from the verification repro, so reviewers can inspect the observability change independently.

</details>

## Non-goals

- No change to task status, terminal state, or exit-code semantics.
- No change to the executor spawn path itself.
- Does not touch the shared repro script or the Step 5 verification slice.
- No data migration.

### Before

```mermaid
graph TD
    A["Startup or forced-stop failure"] --> B["Transient executor output"]
    B --> C["Terminal task state"]
    C --> D["Later inspection sees generic failure"]
```

### After

```mermaid
graph TD
    A["Startup or forced-stop failure"] --> B["Diagnostic formatter"]
    B --> C["Durable task output"]
    C --> D["SQLite output reader"]
    D --> E["Later inspection sees concrete failure details"]
```

## Test Plan

- [x] `pnpm run test:all`

## Revert Plan

- Safe to revert? Yes
- Revert command: `git revert <merge commit sha>`
- Post-revert steps: Re-run `pnpm run test:all`.
- Data migration? No

### DIFF
--- a/packages/app/src/main.ts
+++ b/packages/app/src/main.ts
@@ -2959,12 +2959,6 @@ function createEmbeddedTerminalBackendFromConfig(
                       },
                     };
                     logger.error(`[executing-stall] forcing failure for "${task.id}": ${executingError}`, { module: 'db-poll' });
-                    // Snapshot the concrete failure context into durable output
-                    // before the synthetic stall failure overwrites
-                    // task.execution.error with the coarse stall reason. Tasks
-                    // stalled while still launching collapse a real executor
-                    // startup failure; preserve its recent output tail / startup
-                    // stderr so post-mortem retrieval keeps the concrete details.
                     if (persistence) {
                       persistShutdownDiagnostic(task, persistence, {
                         flushPendingOutput: flushTaskOutput,
--- a/packages/app/src/main.ts
+++ b/packages/app/src/main.ts
@@ -2933,6 +2933,21 @@ function createEmbeddedTerminalBackendFromConfig(
                       },
                     };
                     logger.error(`[executing-stall] forcing failure for "${task.id}": ${executingError}`, { module: 'db-poll' });
+                    // Snapshot the concrete failure context into durable output
+                    // before the synthetic stall failure overwrites
+                    // task.execution.error with the coarse stall reason. Tasks
+                    // stalled while still launching collapse a real executor
+                    // startup failure; preserve its recent output tail / startup
+                    // stderr so post-mortem retrieval keeps the concrete details.
+                    if (persistence) {
+                      persistShutdownDiagnostic(task, persistence, {
+                        flushPendingOutput: flushTaskOutput,
+                        forcedStopReason: executingError,
+                        label: task.execution.phase === 'launching'
+                          ? 'Startup Failure Diagnostic'
+                          : 'Shutdown Diagnostic',
+                      });
+                    }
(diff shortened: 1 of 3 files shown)


EdbertChan:
## Implementation rationale (moved out of inline code comments)

Recording the explanatory comments here so the code stays lean while the reasoning remains discoverable.

### `packages/app/src/main.ts` — forced executing-stall failure path
Snapshot the concrete failure context into durable output before the synthetic stall failure overwrites `task.execution.error` with the coarse stall reason. Tasks stalled while still launching collapse a real executor startup failure; preserve its recent output tail / startup stderr so post-mortem retrieval keeps the concrete details.

### `packages/data-store/src/sqlite-adapter.ts`

**`OUTPUT_DIAGNOSTIC_TAIL_CHARS` constant** — Max characters of the durable diagnostic file (`full/*.log`) folded into a tail read. Bounds the memory footprint of `getOutputTail` when a task has a large diagnostic file while still surfacing concrete startup/shutdown failure details to live-tail and post-mortem snapshots.

**`getOutputTail` — diagnostic fold-in** — Fold in the durable diagnostic file (`full/*.log`). Executor startup failures append their concrete startup stderr/message there via `appendTaskOutput`, and that content never reaches the output spool. A tail read for a synthetic owner-shutdown or startup-stall failure would otherwise snapshot an empty tail and lose the concrete startup context. Append it as a single trailing chunk so both the live tail and post-mortem diagnostic snapshots surface it. Read fresh each call (never cached) so it stays current as diagnostics are appended at terminal time, and bound the read so a large diagnostic file can't blow up the tail's memory footprint.

**`readDiagnosticTail`** — Read at most `maxChars` from the end of the durable diagnostic file (`full/*.log`). Returns `''` when the file is missing or empty. A leading marker is added when the file is truncated so readers know earlier diagnostic content exists.

### `packages/data-store/src/__tests__/sqlite-adapter.test.ts`

**"folds durable diagnostic content into the output tail"** — Executor startup failures append concrete startup stderr via `appendTaskOutput`, which lands only in the diagnostic file (never the spool). `getOutputTail` — the source `persistShutdownDiagnostic` snapshots and the live-tail IPC reads — must surface it so a synthetic owner-shutdown / startup-stall failure retains the concrete context instead of a coarse terminal error alone.

**"returns a startup diagnostic tail even with no spool output"** — An executor that dies before streaming anything leaves only the diagnostic file; the tail must still surface its concrete message.

**Large-output prune assertion** — Expect 6 chunks: 5 spool chunks (bounded by `outputTailLimit`) plus a single bounded diagnostic chunk folded in from the large `full/*.log` file. The diagnostic chunk must be bounded (`< 64 KiB`), not the multi-MB raw file.

PR #1400, opened 2026-06-16
Recovery lifecycle 2b: common worker runtime

EdbertChan:
Validation passes with no warnings. Here is the final PR body:

## Summary

Recovery work now runs as an explicit, long-running worker process instead of hiding inside a GUI session or a one-off headless command.

This adds a shared worker runtime that auto-fix and external recovery will both use. It owns only the mechanics: when to wake, how to avoid overlapping cycles, and shutdown.

The runtime stays policy-free. It wakes on matching workflow lifecycle events, scans for work, and hands each item to a caller-supplied callback. It does not decide what counts as recovery.

A periodic poll plus a startup scan is the safety net. Even if every wakeup event is missed, the worker still makes forward progress and clears any backlog.

You start one headlessly with `worker <autofix|external-recovery|all>`. The headless client runs it locally rather than handing it to a shared mutation owner, so the worker is independently owned and stopped.

`external-recovery` is a no-op runtime for now; its real worker arrives in a later slice. Poll cadence is tunable through the new `workerPollIntervalMs` config value.

### Before

```mermaid
graph TD
    Cmd[Headless command / GUI session] --> Sub[Per-command lifecycle subscription]
    Sub --> Work[Recovery work runs while command stays alive]
    Work -. stops when command exits .-> Gone[No worker left]
```

### After

```mermaid
graph TD
    Start[worker autofix / external-recovery / all] --> RT[Worker runtime process]
    Bus[WORKFLOW_LIFECYCLE events] --> RT
    Poll[Startup scan + periodic poll] --> RT
    RT --> Drain[Serialized drain loop]
    Drain --> Scan[scan: discover work]
    Scan --> Submit[submit: act on each item]
    Signal[SIGINT / SIGTERM] --> Stop[Clean shutdown]
    Stop --> RT
```

## Test Plan

- [ ] `cd packages/app && pnpm test -- src/__tests__/worker-runtime.test.ts src/__tests__/headless-client.test.ts`
- [ ] `pnpm run test:all`

## Revert Plan

- Safe to revert? Yes
- Revert command: `git revert <merge-sha>`
- Post-revert steps: None. The change is additive — a new runtime file, headless `worker` routing, and one config field. No callers depend on it yet.
- Data migration? No

### DIFF
--- a/packages/app/src/__tests__/headless-client.test.ts
+++ b/packages/app/src/__tests__/headless-client.test.ts
@@ -435,6 +435,28 @@ describe('headless-client', () => {
     expect(secondExecHandler).toHaveBeenCalledTimes(1);
   }, 15_000);
 
+  it('routes a worker command directly to the host runtime, never delegating', async () => {
+    const bus = new LocalBus();
+    const execHandler = vi.fn(async () => ({ ok: true }));
+    // Even if a standalone owner is reachable, `worker` must run in-process.
+    bus.onRequest('headless.exec', execHandler);
+    bus.onRequest('headless.owner-ping', async () => ({ ok: true, ownerId: 'owner-1', mode: 'standalone' }));
+
+    const runElectronHeadless = vi.fn(async () => 0);
+    const ensureStandaloneOwner = vi.fn(async () => {});
+
+    const exitCode = await runHeadlessClientCommand(['worker', 'autofix'], {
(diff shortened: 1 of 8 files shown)


EdbertChan:
Closing as superseded by the later Recovery Lifecycle stack that already landed on master.

PR #1452, opened 2026-06-17
Recovery lifecycle 2c: auto-fix worker

EdbertChan:
Validation passes with no warnings. Here is the final PR body:

## Summary

Auto-fix now runs as its own start/stop worker instead of firing from hidden callbacks. You launch it with `./run.sh --headless worker autofix`.

Lifecycle events wake it up, then it reads the saved workflow and task state to decide what needs fixing. The event is only a nudge, not the source of truth.

For an eligible failed task, or a failed review-gate CI run, it submits a normal `fix <taskId> <agent> --auto-fix` command. The same path a person's "Fix with AI" uses.

The worker owns policy only. It never calls the fix execution functions directly, so retry accounting and the actual fix route stay in one place: the accepted-command boundary.

If no owner is reachable, the worker skips and tries again later through a saved intent fallback or a later scan. Polling also recovers any lifecycle event it missed.

## Architecture

This moves failed-task and review-gate CI recovery out of scattered callbacks and into one explicit worker process, changing how recovery is triggered and routed.

### Before

```mermaid
graph TD
    F[Task fails / CI fails] --> CB[TaskRunner review-gate callback]
    F --> SUB[GUI / headless failed-delta subscription]
    CB --> FIX[fixWithAgent / resolveConflict called directly]
    SUB --> FIX
    FIX --> RUN[Fix runs]
```

### After

```mermaid
graph TD
    F[Task fails / CI fails] --> EVT[Workflow lifecycle event]
    EVT --> W[Auto-fix worker wakes]
    POLL[Poll fallback] --> W
    W --> SCAN[Scan persisted state for eligible work]
    SCAN --> CMD[Submit normal fix --auto-fix command]
    CMD --> BND[Accepted-command boundary]
    BND --> RUN[Fix with AI route]
```

## Test Plan

- [ ] `cd packages/app && pnpm test -- src/__tests__/autofix-worker.test.ts src/__tests__/task-runner-wiring.test.ts src/__tests__/workflow-actions.test.ts`
- [ ] `pnpm run test:all`

## Revert Plan

- Safe to revert? Yes
- Revert command: `git revert <merge-sha>` (or drop this stacked branch)
- Post-revert steps: None. Auto-fix returns to the prior callback-driven path on the base branch.
- Data migration? No

### DIFF
--- a/packages/app/src/__tests__/autofix-worker.test.ts
+++ b/packages/app/src/__tests__/autofix-worker.test.ts
@@ -1,0 +1,341 @@
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { Channels, LocalBus } from '@invoker/transport';
+import type { TaskState } from '@invoker/workflow-core';
+import type { WorkflowMutationIntent } from '@invoker/data-store';
+
+import {
+  buildReviewGateCiContext,
+  isReviewGateCiCandidateEligible,
+  scanAutoFixCandidates,
+  startAutoFixWorker,
+  type AutoFixCandidate,
+  type AutoFixWorkerStateView,
+} from '../autofix-worker.js';
+import { buildReviewGateCiFailedLifecycleEvent } from '../lifecycle-events.js';
+
+const logger = {
+  debug: vi.fn(),
+  info: vi.fn(),
+  warn: vi.fn(),
+  error: vi.fn(),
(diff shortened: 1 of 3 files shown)


EdbertChan:
Closing as superseded by the later Recovery Lifecycle stack that already landed on master.

PR #1471, opened 2026-06-18
Fix intent cancellation Step 5: repro and regression verification (restacked v3)

EdbertChan:
Validation passed. Here is the final PR body:

## Summary

This step proves the intent-cancellation fix actually works. It adds no product code. It only runs the proof tests for the fix that earlier steps in this stack landed.

The earlier steps fixed two timing bugs. One let a cancelled workflow apply a stale fix. The other let stale SSH worktree metadata leak into a later run.

This step re-runs the shared reproduction script in "fixed" mode. A passing run means both race conditions are closed end to end, not just patched in theory.

It then runs two focused tests. One guards the app coordinator's workflow-mutation invalidation path. One guards the execution-engine's SSH worktree metadata path.

Finally it runs the full repository test suite. That is the last gate, confirming nothing across packages regressed before this stack merges.

All four checks passed.

## Test Plan

- [ ] `bash scripts/repro/repro-fix-intent-cancellation-and-stale-ssh-metadata.sh --expect fixed`
- [ ] `cd packages/app && pnpm exec vitest run src/__tests__/persisted-workflow-mutation-coordinator.test.ts`
- [ ] `cd packages/execution-engine && pnpm test -- src/__tests__/ssh-worktree-metadata-repro.test.ts`
- [ ] `pnpm run test:all`

## Revert Plan

- Safe to revert? Yes
- Post-revert steps: None — this step adds no product code, only verification runs
- Data migration? No


EdbertChan:

PR #1459, opened 2026-06-18
fix(launch): make outbox dispatch always active

EdbertChan:
## Summary

The launch outbox is now always active.

The old environment flag was a rollout switch. It kept disabled and observe paths alive after the active path became the default.

This removes that switch and deletes the old launch paths. Task starts now always go through the durable dispatch queue.

### Before

```mermaid
flowchart TD
  Config[INVOKER_LAUNCH_OUTBOX] --> Mode{mode}
  Mode --> Disabled[direct TaskRunner launch]
  Mode --> Observe[write rows, direct launch]
  Mode --> Active[LaunchDispatcher owns launch]
```

### After

```mermaid
flowchart TD
  Scheduler[Orchestrator drainScheduler] --> Row[task_launch_dispatch row]
  Row --> Dispatcher[LaunchDispatcher poll]
  Dispatcher --> Runner[owner TaskRunner]
```

## Test Plan

- [x] `pnpm run build` at commit `8efdbc3`
- [x] `scripts/repro/repro-launch-outbox-active-only.sh` at commit `8efdbc3`

## Revert Plan

- Safe to revert? Yes
- Revert command: `git revert 8efdbc353cf1aecefe363fdeb99cea532dd135afc6`
- Post-revert steps: Re-check any local use of the old launch-outbox environment flag.
- Data migration? No

PR #1478, opened 2026-06-19
fix(launch): make outbox dispatch always active

EdbertChan:


EdbertChan:
Closing: this was created from a temporary landing worktree. Updating #1459 instead.

PR #1551, opened 2026-06-20
Clear pool routing when switching to worktree

EdbertChan:
## Summary

This makes "switch to worktree" actually stop using SSH.

The task kept old SSH routing data after the switch. So even though the UI said worktree, the runner could still send the task back to SSH.

The fix clears the leftover SSH routing data when the task moves to worktree.

### Before

```mermaid
flowchart LR
  A[editTaskType to worktree] --> B[Clear poolMemberId only]
  B --> C[poolId still set]
  C --> D[TaskRunner keeps SSH pool routing]
```

### After

```mermaid
flowchart LR
  A[editTaskType to worktree] --> B[Clear poolMemberId and poolId]
  B --> C[No SSH placement remains]
  C --> D[TaskRunner selects worktree]
```

## Test Plan

- [ ] pnpm --filter @invoker/workflow-core exec vitest run src/__tests__/orchestrator.test.ts -t "clears pool placement|switching ssh → worktree"
- [ ] pnpm run check:types

## Revert Plan

- Safe to revert? Yes
- Revert command: `git revert 44d6cc335`
- Post-revert steps: None
- Data migration? No

### DIFF
--- a/packages/workflow-core/src/__tests__/orchestrator.test.ts
+++ b/packages/workflow-core/src/__tests__/orchestrator.test.ts
@@ -4091,7 +4091,7 @@ describe('Orchestrator', () => {
       expect(persisted?.task.config.poolMemberId).toBe('remote_digital_ocean');
     });
 
-    it('clears poolMemberId when switching away from ssh', () => {
+    it('clears pool placement when switching away from ssh', () => {
       orchestrator.loadPlan({
         name: 'edit-type-clear-remote',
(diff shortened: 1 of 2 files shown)


EdbertChan:
Closing this as obsolete.

This PR cleared pool routing when a task was switched directly to worktree. That keeps the bad API alive.

The replacement is #1555: tasks are assigned to pools, and executors stay internal to those pools.

PR #1561, opened 2026-06-20
Remove core direct executor mutation

EdbertChan:
## Summary

This removes the core direct executor mutation API.

The orchestrator and command service no longer expose `editTaskType`. Core task routing now keeps executor choice behind pool scheduling.

Tests that only covered direct executor edits were removed or switched to pool assignment.

### Before

```mermaid
flowchart LR
  CommandService --> EditType[editTaskType]
  EditType --> Orchestrator
  Orchestrator --> RunnerKind[task.execution.runnerKind]
```

### After

```mermaid
flowchart LR
  CommandService --> EditPool[editTaskPool]
  EditPool --> Orchestrator
  Orchestrator --> Pool[task.execution.poolId]
  Pool --> Scheduler[executor chosen internally]
```

## Test Plan

- [x] pnpm --filter @invoker/workflow-core exec vitest run src/__tests__/command-service.test.ts src/__tests__/orchestrator.test.ts src/__tests__/orchestrator-gates-and-workflow-admin.test.ts src/__tests__/state-topology-matrix.test.ts

## Revert Plan

- Safe to revert? Yes
- Revert command: `git revert 8a478e2cb
- Post-revert steps: None
- Data migration? No

Depends-On: #1560

PR #1562, opened 2026-06-20
Enforce TS-safe imports for source-export packages

EdbertChan:
## Summary

This adds a guard for TypeScript source-export packages.

Those packages must not use relative `.js` imports into another package's source. The check catches that before CI accepts the stack.

The required test suite now runs the guard with the rest of the repository checks.

## Test Plan

- [x] pnpm run check:types
- [x] pnpm run check:ts-source-imports

## Revert Plan

- Safe to revert? Yes
- Revert command: `git revert 61cd33392
- Post-revert steps: None
- Data migration? No


EdbertChan:
Held back from this Mergify landing stack. Current landing top is #1561.

PR #1720, opened 2026-06-21
Register standalone owner handlers before launch-dispatch polling

Siddhant-K-code:
## Summary

Closes INV-192.

A standalone owner could make IPC reachable before registering its owner identity handler.

This patch starts launch-dispatch polling after the owner identity, exec, and GUI mutation handlers are registered.

The repro script demonstrates the bad race window with the real IPC transport: IPC can be reachable while `headless.owner-ping` still returns `NO_HANDLER` before registration.

The current production check is a source-order guard. The deeper boot-level check is tracked separately in INV-222.

<details>

Review Claim: Approve the narrow startup-order patch that prevents dispatch polling from preceding owner handler registration.

Review Lane: behavior

Review Unit: activation-surface

Safety Invariant: The change only reorders standalone owner startup within the same block; read-only mode still skips launch dispatch.

Slice Rationale: This patch closes the observed handler-registration gap while the broader runtime proof is tracked independently.

</details>

## Non-goals

- Does not change delegated routing semantics.
- Does not change launch-dispatch polling behavior once started.
- Does not add the boot-level owner handler check; that follow-up is tracked in INV-222.

### Before

```mermaid
graph TD
    A["standalone owner starts"] --> B["launch-dispatch polling can start"]
    B --> C["owner handlers registered later"]
```

### After

```mermaid
graph TD
    A["standalone owner starts"] --> B["owner and exec handlers registered"]
    B --> C["launch-dispatch polling starts"]
```

## Test Plan

- [x] `scripts/repro/repro-standalone-owner-ping-before-dispatch.sh`
- [x] `pnpm --filter @invoker/app exec vitest run src/__tests__/standalone-owner-handler-ordering.test.ts`
- [x] `pnpm run check:types`

## Revert Plan

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

## Release Notes

* **Bug Fixes**
  * Improved standalone headless startup sequencing so the standalone owner registers its IPC request handlers before the dispatcher begins polling, reducing the risk of “no handler” connection failures during startup.

* **Tests**
  * Added a regression test that verifies the ordering of the standalone owner IPC handler setup relative to dispatcher startup to prevent future reordering issues.

### DIFF
--- a/packages/app/src/__tests__/standalone-owner-handler-ordering.test.ts
+++ b/packages/app/src/__tests__/standalone-owner-handler-ordering.test.ts
@@ -1,0 +1,35 @@
+import { describe, expect, it } from 'vitest';
+import { readFileSync } from 'node:fs';
+import * as path from 'node:path';
+
+const MAIN = path.resolve(__dirname, '..', 'main.ts');
+
+/**
+ * Regression guard for INV-192: a standalone owner must register its owner
+ * discovery (headless.owner-ping) and exec handlers before it starts
+ * launch-dispatch polling. If polling starts first, the owner can own the IPC
+ * socket while owner-ping still returns NO_HANDLER, so peer clients connect but
+ * cannot resolve a standalone owner and mutating commands fail closed.
+ */
+describe('standalone owner handler ordering', () => {
+  it('registers owner discovery and exec handlers before starting launch-dispatch polling', () => {
+    const source = readFileSync(MAIN, 'utf8');
+
+    const ownerPingIdx = source.indexOf("messageBus.onRequest('headless.owner-ping'");
+    const execIdx = source.indexOf("messageBus.onRequest('headless.exec'");
+    const dispatcherIdx = source.indexOf('startStandaloneLaunchDispatcher({');
+
+    expect(ownerPingIdx, 'standalone headless.owner-ping handler not found').toBeGreaterThan(-1);
+    expect(execIdx, 'standalone headless.exec handler not found').toBeGreaterThan(-1);
+    expect(dispatcherIdx, 'startStandaloneLaunchDispatcher call not found').toBeGreaterThan(-1);
+
+    expect(
+      ownerPingIdx,
+      'INV-192: startStandaloneLaunchDispatcher must run after headless.owner-ping is registered',
+    ).toBeLessThan(dispatcherIdx);
+    expect(
+      execIdx,
+      'INV-192: startStandaloneLaunchDispatcher must run after headless.exec is registered',
+    ).toBeLessThan(dispatcherIdx);
+  });
+});
--- a/packages/app/src/__tests__/standalone-owner-handler-ordering.test.ts
+++ b/packages/app/src/__tests__/standalone-owner-handler-ordering.test.ts
@@ -4,13 +4,6 @@ import * as path from 'node:path';
 
 const MAIN = path.resolve(__dirname, '..', 'main.ts');
 
-/**
- * Regression guard for INV-192: a standalone owner must register its owner
- * discovery (headless.owner-ping) and exec handlers before it starts
- * launch-dispatch polling. If polling starts first, the owner can own the IPC
- * socket while owner-ping still returns NO_HANDLER, so peer clients connect but
- * cannot resolve a standalone owner and mutating commands fail closed.
- */
 describe('standalone owner handler ordering', () => {
   it('registers owner discovery and exec handlers before starting launch-dispatch polling', () => {
     const source = readFileSync(MAIN, 'utf8');
--- a/packages/app/src/main.ts
+++ b/packages/app/src/main.ts
@@ -1229,14 +1229,6 @@ function startHeadlessMode(): void {
 
       // In standalone owner mode, serve delegated requests from peer headless processes.
       if (standaloneMode && messageBus) {
-        if (!readOnlyMode) {
-          standaloneLaunchDispatcherController = startStandaloneLaunchDispatcher({
-            headlessDeps,
-            ownerId: workflowMutationOwnerId,
-            createTaskExecutor: createStandaloneTaskExecutor,
-            setLatestTaskExecutor: (executor) => { latestTaskExecutor = executor; },
-          });
-        }
         const standaloneOwnerIdleTimeoutMs = Number.parseInt(
           process.env.INVOKER_STANDALONE_OWNER_IDLE_TIMEOUT_MS ?? '0',
           10,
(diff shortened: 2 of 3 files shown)


Siddhant-K-code:
CI note: the only red check I found is `inv117 / deterministic proof`, failing in `required/11-pr-authoring-guardrails.sh`. This appears inherited from the base, not introduced by this handler-ordering diff: `validate-pr-body.mjs` requires `## Review Unit`, while `scripts/test-pr-body-validator.mjs` still has “valid” fixtures without that section on `origin/master` and this branch. The PR remains blocked until that shared guardrail fixture mismatch is fixed or the branch rebases onto a base that has it fixed.

packages/app/src/__tests__/standalone-owner-handler-ordering.test.ts
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Regression guard for INV-192: a standalone owner must register its owner
+ * discovery (headless.owner-ping) and exec handlers before it starts
+ * launch-dispatch polling. If polling starts first, the owner can own the IPC
+ * socket while owner-ping still returns NO_HANDLER, so peer clients connect but
+ * cannot resolve a standalone owner and mutating commands fail closed.
+ */
+describe('standalone owner handler ordering', () => {
+  it('registers owner discovery and exec handlers before starting launch-dispatch polling', () => {
+    const source = readFileSync(MAIN, 'utf8');
```
Siddhant-K-code, line 9: The fix is the right call — registering owner-ping/exec before starting dispatch closes the window the issue describes. On the guard itself: it asserts textual order of three string literals in main.ts, which follows the `no-manual-dispatch` precedent, but it doesn't exercise the actual race INV-192 calls out (owner-ping resolving before active dispatch polling). It would still pass if a later change reintroduced the race a different way — e.g. a setup-level `await` slipped in between the socket going live and the owner-ping registration, or the socket bind moving earlier — since none of that changes the textual order of these literals.

INV-192 specifically asks for "a regression test that a bootstrapped standalone owner responds to owner discovery before active dispatch polling can run." Worth a short behavioral test (boot a standalone owner, assert `headless.owner-ping` resolves before the dispatcher's first poll), or at minimum a comment noting this guard only protects source order. Non-blocking.

Siddhant-K-code: I am keeping this as non-blocking. The branch already has the narrow source-order guard in `standalone-owner-handler-ordering.test.ts` and the main-process comment explaining why dispatch starts only after `headless.owner-ping` and exec handlers are registered. A full boot-level owner-ping-before-first-poll harness would be broader than this ordering fix.

Siddhant-K-code:
CI update: this is no longer blocked by the deterministic-proof failure noted above. The current run is green, including `inv117 / deterministic proof`.

Siddhant-K-code:
@coderabbitai review

packages/app/src/main.ts
```diff
@@ -1705,6 +1697,19 @@ function startHeadlessMode(): void {
           getTaskExecutor: createStandaloneTaskExecutor,
           logger,
         });
+
+        // Start launch-dispatch polling only after the owner discovery (headless.owner-ping)
```
EdbertChan: Excessive comments, these should be in the Github pr

packages/app/src/__tests__/standalone-owner-handler-ordering.test.ts
```diff
@@ -0,0 +1,35 @@
...
+
+const MAIN = path.resolve(__dirname, '..', 'main.ts');
+
+/**
+ * Regression guard for INV-192: a standalone owner must register its owner
+ * discovery (headless.owner-ping) and exec handlers before it starts
+ * launch-dispatch polling. If polling starts first, the owner can own the IPC
+ * socket while owner-ping still returns NO_HANDLER, so peer clients connect but
+ * cannot resolve a standalone owner and mutating commands fail closed.
+ */
```
EdbertChan: Ibid. Comments like these should be jsutification for the PR. in fact, that should probably be in the pr summary.

When commenting and deciding where to comment, you're talking to a future person.  if its in the code, the comment means that it is tightly coupled to the architectural invariants.

If its in the pr summary, its a justification for the change

if its in the pr comment, its an anticipation of the reviewer's concerns.

packages/app/src/__tests__/standalone-owner-handler-ordering.test.ts
EdbertChan: If this can be reintroduced and your fix is just a patch, please declare it as such in the PR Summary and file a ticket in Linear for the real fix/investigation.

EdbertChan (dismissed):
Just move the comments out and put them in the Pr summary or Github comments

packages/app/src/__tests__/standalone-owner-handler-ordering.test.ts
Siddhant-K-code: Addressed in d094feb4. I moved the patch limitation and justification into the PR summary, trimmed the code/test comments down to the local invariant, and filed INV-222 for the boot-level owner discovery proof.

packages/app/src/main.ts
Siddhant-K-code: Addressed in d094feb4. I removed the review-style explanation from code and left only the local ordering invariant.

packages/app/src/__tests__/standalone-owner-handler-ordering.test.ts
Siddhant-K-code: Addressed in d094feb4. I removed the review-style explanation from code and left only the local ordering invariant.

Siddhant-K-code:
Addressed the review feedback in the latest push: removed the long explanatory comments from the diff, kept only the compact ordering invariant in code, moved the limitation/follow-up context into the PR summary, and rebased the branch onto current master.

Siddhant-K-code:
@coderabbitai review

EdbertChan (approved):
@mergifyio queue

EdbertChan (dismissed):
This is a theoretical issue and I'm not fully sure I understand. it looks like you moved the dispatcher into a gui-exec event?

Siddhant-K-code:
Addressed with `5f4bf76f`.

I added `scripts/repro/repro-standalone-owner-ping-before-dispatch.sh` so this is no longer only a theoretical/source-order claim. The script uses the real IPC transport to show the bad window: the socket is reachable and a dispatch poll can run while `headless.owner-ping` still returns `NO_HANDLER` before owner handler registration. After registration, the same request succeeds.

Also clarifying the control-flow concern: this does not move launch dispatch into `headless.gui-mutation`. Dispatch polling now starts after the standalone owner registers `headless.owner-ping`, `headless.exec`, and `headless.gui-mutation`, so the owner activation surface exists before polling begins.

Validated with:
- `scripts/repro/repro-standalone-owner-ping-before-dispatch.sh`
- `pnpm --filter @invoker/app exec vitest run src/__tests__/standalone-owner-handler-ordering.test.ts`
- `pnpm run check:types`

Siddhant-K-code:
@coderabbitai review

Siddhant-K-code:
Addressed in d115edc4. I shortened the repro temp-directory template from `invoker-owner-ping-before-dispatch.XXXXXX` to `inv-ping.XXXXXX`, keeping `ipc.sock` well under Unix-domain socket path limits when `$TMPDIR` is long.

Verification:
- `git diff --check`
- `PATH=/tmp/invoker-bin:/tmp/node-v26.3.1/bin:$PATH scripts/repro/repro-standalone-owner-ping-before-dispatch.sh`

EdbertChan (approved):
In the future, keep explainations at a high level. What matters is the flow and architecture of the code and how these components are communicating.

The structure enforces that. We're effectively saying the logic and communication is incorrect. So let's operate on that level

If you tell the AI to keep responses and explanations at a high level and < 40 words it does a good job at that

Siddhant-K-code:
got it!!

Siddhant-K-code:
@mergifyio queue

PR #1724, opened 2026-06-21
Improve planning bootstrap and skill diagnostics

Siddhant-K-code:
Fixes INV-205

Superseded by the Mergify stack:
- #1822
- #1823
- #1824

Introduced:
- GUI owner bootstrap fallback from daemon startup timeout to local owner mode.
- Clear bundled-skill diagnostics for missing vs stale managed skill targets.
- Planning prompt coverage for concise assumptions/preview on small or ambiguous requests.
- Plan-to-Invoker verification-command discovery support for repos without `test:all`.

Reduced:
- Startup dead-end behavior when daemon ownership is unavailable.
- Ambiguous skill install state in setup UI.
- Hardcoded final verification assumptions in generated implementation plans.

Tests:
- pnpm --filter @invoker/app exec vitest run src/__tests__/app-bootstrap.test.ts src/__tests__/bundled-skills.test.ts
- pnpm --filter @invoker/surfaces exec vitest run src/__tests__/plan-conversation.test.ts
- pnpm run test:plan-to-invoker-skill
- pnpm run check:types
- pnpm run check:all


packages/app/src/bootstrap/app-bootstrap.ts
```diff
@@ -1,5 +1,34 @@
...
 
+export type GuiOwnerPreference = 'auto' | 'daemon' | 'gui';
+
+export function resolveGuiOwnerPreference(env: NodeJS.ProcessEnv = process.env): GuiOwnerPreference {
+  const raw = (env.INVOKER_GUI_OWNER_MODE ?? '').trim().toLowerCase();
+  if (raw === 'daemon' || raw === 'client' || raw === 'follower') return 'daemon';
+  if (raw === 'auto') return 'auto';
+  if (raw === 'gui' || raw === 'owner' || raw === 'local') return 'gui';
+  if (env.INVOKER_GUI_DAEMON_OWNER === '1') return 'daemon';
+  return 'auto';
```
Siddhant-K-code, line 11: Flipping the default to `auto` here has a side effect worth confirming. When `auto` discovers a running daemon owner, the GUI comes up as a read-only client (`ownerMode = false` in `runGuiReadyBootstrap`) and delegates mutations to that owner. The delegation path in `ipc/ipc-registration.ts` recovers from `NO_HANDLER`/`DISCONNECTED` by calling `refreshOwnerRoute()` and retrying — but `refreshGuiMutationOwnerRoute()` (main.ts:511) early-returns unless the preference is exactly `'daemon'`. So in the new default, if the owner restarts or the bus drops mid-session, the refresh no-ops and the retry hits the same dead bus → "No mutation owner is available", with no re-discovery or rebuild.

Before this change the shipped default (`daemon`) exercised that recovery; now it's effectively disabled for the common case. Can we extend the guard to also handle `'auto'` when running as a client — re-discover the owner and rebuild the bus (without spawning) — so auto-mode clients don't silently lose owner-route recovery?

skills/plan-to-invoker/scripts/lint-task-atomicity.sh
```diff
@@ -710,9 +711,9 @@ END {
     final_command = task_command_line[final_idx]
 
     if (task_has_command[final_idx] != 1) {
-      errors[++errn] = "Task \"" final_id "\" must be a command task because standalone implementation plans and terminal stack workflows require a final pnpm run test:all regression gate"
-    } else if (final_command != "pnpm run test:all") {
-      errors[++errn] = "Task \"" final_id "\" must be the final full-suite regression gate and run exactly \"pnpm run test:all\""
+      errors[++errn] = "Task \"" final_id "\" must be a command task because standalone implementation plans and terminal stack workflows require a final discovered regression gate"
+    } else if (final_command != "pnpm run test:all" && task_has_verification_discovery[final_idx] != 1) {
```
Siddhant-K-code: Non-blocking: this passes any non-`test:all` final command as long as the task description contains the literal phrase "Verification command discovery:". The flag at line 463 only matches that phrase — it doesn't check that `final_command` is the discovered command, so a plan could ship a no-op final gate (e.g. `true`) plus the phrase and still lint clean. Worth tying the pass to the discovered command actually being referenced, or at least asserting it's non-trivial, so the external-repo escape hatch can't be used to skip a real regression gate.

packages/app/src/bootstrap/app-bootstrap.ts
Siddhant-K-code: Addressed in bd5dbcb8. `auto` GUI clients now track when they are daemon-backed and refresh/rebuild the owner route after rediscovery without spawning a daemon; local GUI owners still skip this path.

skills/plan-to-invoker/scripts/lint-task-atomicity.sh
Siddhant-K-code: Addressed in bd5dbcb8. Non-`pnpm run test:all` final gates now reject trivial commands and require the `Verification command discovery:` text to reference the final command, with fixture coverage for both cases.

Siddhant-K-code:
Superseded by the Mergify stack for easier review:

- #1822
- #1823
- #1824

PR #2156, opened 2026-06-23
[Fix intent cancellation Step 6: worker response lineage guard (restacked v3)](1) Require worker responses to match both selectedAttemptId and executionGeneration

EdbertChan:
## Summary

The orchestrator routes each worker response to the task attempt that asked for it. This change tightens which responses it will accept.

A worker response is now honored only when it carries the currently selected attempt id and the current execution generation together.

If a response arrives with the right attempt id but an out-of-date generation, the orchestrator drops it instead of routing it through.

This keeps a leftover reply from a superseded run from landing on a task that has already advanced.

<details>

Review Claim: A worker response is honored only when it carries the live attempt id and the live execution generation together.

Review Lane: behavior

Review Unit: routing

Safety Invariant: The change only narrows which worker responses the orchestrator routes; older replies that omit a field keep their prior handling, so nothing new is accepted.

Slice Rationale: This slice carries one orchestrator routing guard and nothing else, so it can be reviewed on its own.

</details>

## Non-goals

This does not change how attempts are started, how generations are assigned, or any persistence behavior.

## Test Plan

## Revert Plan

- Safe to revert? Yes
- Revert command: `git revert 80ce3087db6b623159f4b690e1180587f11709f1`
- Post-revert steps: None
- Data migration? No

### DIFF
--- a/packages/workflow-core/src/__tests__/orchestrator.test.ts
+++ b/packages/workflow-core/src/__tests__/orchestrator.test.ts
@@ -855,6 +855,62 @@ describe('Orchestrator', () => {
       }
     });
 
+    it('rejects a completion signal when the current attemptId carries a stale executionGeneration', () => {
+      const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
+      try {
+        orchestrator.loadPlan({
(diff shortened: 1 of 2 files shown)


EdbertChan:

PR #2160, opened 2026-06-23
Fix intent cancellation Step 6: worker response lineage guard (restacked v3)

EdbertChan:
## Summary

Worker responses now have to prove they belong to the task's current run before the orchestrator accepts their output.

Before, a response that carried the live attempt id but a stale execution generation slipped through. The lineage check only ran when the attempt id was missing.

Now the orchestrator validates whichever lineage fields the producer sent. When both the attempt id and the execution generation are present, both must match the live task.

This closes Step 6 of the intent-cancellation fix: a late response from a superseded run can no longer overwrite the work of the current run.

<details>

Review Claim: A worker response carrying the current attempt id but a stale execution generation is rejected.

Review Lane: `behavior`

Review Unit: `routing`

Safety Invariant: The guard only tightens acceptance; responses that already matched both fields still pass, and producers that omit a field keep their old leniency.

Slice Rationale: This is the single orchestrator guard for Step 6, kept separate from the earlier cancellation steps so the lineage check reviews on its own.

</details>

## Non-goals

- No change to how execution generations are assigned or bumped.
- No change to producers that omit the attempt id or the generation.
- No change to cancellation Steps 1 through 5.

## Test Plan

## Revert Plan

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

## Bug Fixes

* Enhanced workflow execution handling to properly reject stale generation responses, preventing outdated execution states from being incorrectly processed even when the attempt ID matches the current selection.

### DIFF
--- a/packages/workflow-core/src/__tests__/orchestrator.test.ts
+++ b/packages/workflow-core/src/__tests__/orchestrator.test.ts
@@ -855,6 +855,62 @@ describe('Orchestrator', () => {
       }
     });
 
+    it('rejects a completion signal when the current attemptId carries a stale executionGeneration', () => {
+      const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
+      try {
+        orchestrator.loadPlan({
+          name: 'reject-stale-generation-with-current-attempt',
+          onFinish: 'none',
+          tasks: [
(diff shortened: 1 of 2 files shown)


EdbertChan:

PR #2254, opened 2026-06-24
[Fix intent cancellation](6) Fix intent cancellation Step 6: worker response lineage guard (regression gated v2)

EdbertChan:
## Summary

A worker response that carried an attempt id was accepted even when its execution generation no longer matched the live task.

That gap let an answer from an older run land on the current run and overwrite fresh state.

This change makes the orchestrator check the execution generation on its own, independent of the attempt id.

When both identity fields are present, both must match the live task before the answer is honored.

Older producers may send only one field, so each field is checked only when present. When both are present, both must agree.

<details>

Review Claim: Approve guarding worker-response handling so an answer with a matching attempt id but an old execution generation is dropped instead of honored.

Review Lane: behavior

Review Unit: routing

Safety Invariant: Behavior only tightens which worker responses are honored; a response with no execution generation, or one whose execution generation still matches the live task, is handled exactly as before. No stored data shape changes.

Slice Rationale: This slice touches only the worker-response identity guard in the orchestrator. It is kept apart from the Step 5 base it builds on so reviewers can approve one guard in isolation.

</details>

## Non-goals

- Does not change how attempt ids are assigned or selected.
- Does not change how the execution generation counter is advanced.
- Does not change any persisted record shape or migration.
- Does not touch the merge gate, launch dispatch, or any other step in the series.

## Test Plan

## Revert Plan

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

### DIFF
--- a/packages/workflow-core/src/__tests__/orchestrator.test.ts
+++ b/packages/workflow-core/src/__tests__/orchestrator.test.ts
@@ -855,6 +855,60 @@ describe('Orchestrator', () => {
       }
     });
 
+    it('rejects a completion signal when attemptId matches but executionGeneration is stale', () => {
+      const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
+      try {
+        orchestrator.loadPlan({
+          name: 'reject-stale-generation-with-current-attempt',
+          onFinish: 'none',
+          tasks: [
+            { id: 'A', description: 'Root', command: 'echo A' },
+          ],
+        });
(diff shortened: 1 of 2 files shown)


EdbertChan:
Closing redundant 'regression gated v2' duplicate. Step 6 (worker response lineage guard) already merged to master via #2164/#1480. Invoker workflow wf-1782192503477-17 cancelled.

PR #2256, opened 2026-06-24
[Recovery Lifecycle 2i Step 0](1) Plan parser execution model field

EdbertChan:
## Summary

Let a plan task pin which model runs it, end to end. Today a plan can choose its harness, but the chosen model never reaches the agent.

Two gaps caused this. The plan readers dropped the model field, and the task store had no column for it, so a saved choice was forgotten.

This change reads the model from a plan task, carries it into the task config beside the harness, and stores it so it survives a save and reload.

It is additive. A plan that omits the model behaves exactly as before.

<details>

Review Claim: A plan task can set executionModel and it is carried through the readers and stored so the chosen model survives a save and reload, mirroring executionAgent.

Review Lane: behavior

Review Unit: routing

Safety Invariant: executionModel is optional and additive everywhere. The new stored column defaults to null, and a task that omits it behaves exactly as before, so no existing plan or stored task changes meaning.

Slice Rationale: This is one cohesive enabling change that makes the model pinnable end to end. It lands first so later slices can declare their model in the plan rather than only at runtime.

</details>

## Non-goals

- Does not change executionAgent behavior.
- Does not change how any harness consumes the model.
- Does not change the orchestrator runtime model command.
- Does not run the full repository regression; that belongs to the 2i terminal slice.

## Test Plan

- [ ] `cd packages/data-store && pnpm test`

## Revert Plan

- Safe to revert? Yes
- Revert command: `git revert <merge-sha>`
- Post-revert steps: None. The stored column is additive and unused once the code is gone.
- Data migration? No. The column defaults to null, so existing rows stay valid after a revert.


EdbertChan:
Closing because the Recovery Lifecycle 2i work has already landed.

PR #2262, opened 2026-06-24
Recovery Lifecycle 2i Step 1: Worker registry contract

EdbertChan:
## Summary

This adds a place to list workers by name inside `@invoker/execution-engine`. The existing auto-fix worker is the first one registered.

Before, the system only knew about one worker, and that knowledge was baked into a single hard-coded branch. Now there is one spot where workers are declared.

Nothing uses this list yet. No entry point reads from it, so how the app runs today does not change at all.

This is the first slice of the 2i work. Later slices will make auto-fix and future workers look up their worker here instead of relying on the one hard-coded path.

## Review Claim

A worker registry exists in `@invoker/execution-engine` and the auto-fix worker is registered as the built-in default; nothing goes through it yet.

## Review Lane

`behavior`

## Safety Invariant

This step is additive and the registry is unused by any entry point, so the auto-fix path and runtime behavior are unchanged from where the 2h stack left them.

## Slice Rationale

The registry is the lowest-layer primitive every later 2i slice builds on, so it lands first and on its own instead of being bundled with the entry-point wiring (step 3) or external worker loading (step 4).

## Non-goals

- Do not wire the registry into the headless or CLI entry points (step 3).
- Do not change the single-instance lock (step 2).
- Do not add external worker loading (step 4).

## Test Plan

- [ ] `grep -R "createWorkerRegistry" packages/execution-engine/src`

## Revert Plan

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No


EdbertChan:
Superseded by #2278, which publishes this slice as a Mergify stack PR targeting `master` (this PR targeted the step-0 branch and used the older PR body format).

PR #2493, opened 2026-06-28
[Detached Workflow Base](1) Prove stale stack base

EdbertChan:
## Summary

This adds a repro for detached workflows that keep an old stack branch as their base.

The case uses `it.fails`, so this proof slice passes before the behavior fix lands.

A small repro script runs just this case for fast local confirmation.

<details>

Review Claim: The managed-base bug is covered by a focused failing regression case.

Review Lane: proof

Review Unit: proof

Safety Invariant: The new case uses `it.fails`, so it proves the current bug without touching runtime code.

Slice Rationale: This lands before the fix so reviewers can see the broken state independently.

</details>

## Non-goals

This does not alter detach behavior, merge behavior, or workflow scheduling.

## Test Plan

- [x] `bash scripts/repro/repro-detached-workflow-stale-stack-base.sh`

## Revert Plan

- Safe to revert? Yes
- Revert command: `git revert 77fbf433a`
- Post-revert steps: None
- Data migration? No

## Summary by CodeRabbit

* **Tests**
  * Added coverage for detaching a workflow when upstream branch metadata is missing, confirming the workflow’s external dependency is cleared and a fallback base branch is used.
* **Documentation**
  * Added a reproducible script to run the specific detachment/fallback scenario in isolation.

### DIFF
--- a/packages/workflow-core/src/__tests__/orchestrator-gates-and-workflow-admin.test.ts
+++ b/packages/workflow-core/src/__tests__/orchestrator-gates-and-workflow-admin.test.ts
@@ -1438,6 +1438,40 @@ describe('Orchestrator', () => {
         },
       ]);
     });
+    it.fails('falls back managed stack base when detached upstream metadata is unavailable', () => {
+      orchestrator.loadPlan({
+        name: 'fallback-upstream',
+        baseBranch: 'master',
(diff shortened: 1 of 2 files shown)


EdbertChan:
Collapsed into #2494 because the repro is the Vitest regression in the final behavior PR.

PR #2536, opened 2026-06-29
[Recovery Lifecycle 2i Step 6: Plan parser execution model field](1) Persist execution model in the task store

EdbertChan:
## Summary

A plan task can now keep its pinned execution model after persistence.

The plan-reader path already carries executionModel on this stack.

This slice makes the task store save and load that same value.

That keeps the pinned model available after a save and reload.

The new column is nullable and optional, so tasks that omit executionModel keep loading with undefined.

<details>

Review Claim: The task store persists config.executionModel on save and load, so a pinned plan model survives reload.

Review Lane: behavior

Review Unit: write-path

Safety Invariant: executionModel is optional and additive; the new stored column defaults to null, so tasks that omit the model keep loading as undefined.

Slice Rationale: Persistence is the durable home for the plan-level selector, kept separate from the parser and worker slices that produce or consume it.

</details>

## Non-goals

- Does not change executionAgent behavior.
- Does not change how plan readers or harnesses choose a model at runtime.
- Does not change the orchestrator runtime model command.
- Does not run the full repository regression; that belongs to the terminal 2i slice.

## Test Plan

- [x] `cd packages/app && pnpm test`
- [x] Round-trip coverage saves a task with `executionModel: "claude"` and reloads it as `claude`; omitted values reload as `undefined`.

## Revert Plan

- Safe to revert? Yes.
- Revert command: `git revert b7f314c934d2b3445031fc28522778c963a020e3`
- Post-revert steps: None.
- Data migration? Additive nullable `execution_model` column only; reverted code ignores the column, so no rollback step is required.

### DIFF
--- a/packages/data-store/src/__tests__/sqlite-adapter.test.ts
+++ b/packages/data-store/src/__tests__/sqlite-adapter.test.ts
@@ -3247,6 +3247,24 @@ describe('SQLiteAdapter', () => {
     });
   });
 
+
+  describe('executionModel persistence', () => {
+    it('round-trips a saved executionModel and leaves omitted values undefined', () => {
+      adapter.saveWorkflow(testWorkflow);
+      const withModel = makeTask('t-model-1', {
+        config: { workflowId: 'wf-1', executionModel: 'claude' },
+      });
+      const withoutModel = makeTask('t-model-2', {
+        config: { workflowId: 'wf-1' },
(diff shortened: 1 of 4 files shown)


EdbertChan:
Closing because the Recovery Lifecycle 2i work has already landed.

PR #3196, opened 2026-07-07
[PR Maintenance Workers](4) Retire temporary PR cron installers

EdbertChan:
## Summary

Renames the old PR cron scripts as worker tick scripts.

These scripts are the actual work units for optional PR maintenance workers. They still do the same jobs, but they are no longer described as host crontab installers.

This PR also removes the temporary cron installer wrappers because the owner worker controller is the new runner.

<details>
<summary>Review metadata</summary>

Review Claim: PR maintenance scripts are now worker-owned tick scripts, not temporary crontab jobs.

Review Lane: policy

Review Unit: tooling-policy

Safety Invariant: The script entrypoints stay in place; only installer wrappers and wording are removed.

Slice Rationale: This cleanup lands before the app starts those scripts through registered workers.

</details>

## Non-goals

No owner worker registration.

No change to CodeRabbit or conflict rebase decision logic.

No live GitHub mutation in tests.

### Before

```mermaid
graph TD
    A["host crontab"] --> B["PR cron scripts"]
```

### After

```mermaid
graph TD
    A["registered worker tick"] --> B["PR maintenance scripts"]
```

## Test Plan

- [x] `bash scripts/repro/repro-pr-crons-mutex.sh`

## Revert Plan

- Safe to revert? Yes
- Revert command: `git revert 5dcdd6ba9`
- Post-revert steps: Reinstall cron wrappers only if returning to crontab-managed execution.
- Data migration? No

PR #3042, opened 2026-07-04
Add timeout guard to master-head cron

EdbertChan:
## Summary

The master-head cron now has a hard test timeout, a working Codex default on DigitalOcean, and cleanup for failed or orphaned child processes.

Before this, a stuck test could hold the cron lock forever. OMP also used an unsupported provider name and could leave its own test child running.

The worker now turns stalls into failed runs, checks OMP auth first, and clears leftover checkout processes even when OMP itself fails or times out.

<details>
<summary>Review metadata</summary>

Review Claim: Approve the master-head cron worker, timeout guard, supported Codex default, and failed-run cleanup.

Review Lane: policy

Review Unit: tooling-policy

Safety Invariant: This only changes cron scripts and their installer test. It does not change product runtime code.

Slice Rationale: This keeps cron reliability separate from any later test-suite fixes the cron discovers.

</details>

## Non-goals

This does not fix the UI test that stalls on DigitalOcean. It makes that stall report cleanly and prevents leaked child runs, including OMP timeout paths.

## Test Plan

- [x] `bash -n scripts/cron-master-head-test-autofix.sh scripts/install-pr-cron-jobs.sh scripts/uninstall-pr-cron-jobs.sh scripts/test-install-pr-cron-jobs.sh`
- [x] `bash scripts/test-install-pr-cron-jobs.sh`
- [x] `omp --no-title --auto-approve --model openai-codex/gpt-5.3-codex-spark -p "Reply with ok only."`

## Revert Plan

- Safe to revert? Yes
- Revert command: `git revert <sha>`
- Post-revert steps: Reinstall cron jobs if this already shipped to a host.
- Data migration? No

## Summary by CodeRabbit

* **New Features**
  * Added a cron-driven “master-head destructive test autofix” workflow that runs tests, optionally confirms failures, and opens a repair PR with visual proof when UI-impacting changes are detected.
  * Expanded maintenance cron install/validation/removal to include the new master-head autofix job with a configurable schedule.
* **Bug Fixes**
  * Improved bounded parallelism and per-item claim locking to reduce overlapping repair attempts.
  * Strengthened timeout handling to terminate entire process groups to prevent leaked descendants.
* **Tests**
  * Added/updated cron job and timeout fallback repro scenarios, including retention and “no-diff” edge cases.

PR #3219, opened 2026-07-07
Save DigitalOcean 1 dangling changes

EdbertChan:
Captured uncommitted DigitalOcean 1 changes before resetting its local master to upstream/master.

The branch was saved first so the remote reset does not lose work. A rebase onto the latest upstream/master conflicted in CHANGELOG.md, slack-surface-workflows.test.ts, plan-conversation.ts, and slack-surface.ts, so this PR keeps the original captured commit for review.


EdbertChan:
Closing: this was a DO1 dangling-changes snapshot from 2026-07-07 that has since been fully superseded. Concretely, of the 25 files: 13 are already applied on master via later PRs (three-way merge is a no-op), 3 Slack files would actually revert landed improvements (admin gating on `exec local:`, per-repo checkout, output-buffer caps, stacked-workflow prompts, stricter agent-thread submit guards), 4 cron-installer files were retired or rewritten by #3196 / the PR-Maintenance-Workers stack, and both CHANGELOG bullets are already on master. The only novel content was two new cron scripts (`cron-admin-bypass-requeue.sh`, `cron-master-head-test-autofix.sh`); we're moving away from host crons toward in-app workers, so those should be reimplemented as workers (or not at all — #3042 was already closed unmerged) rather than salvaged here. See #3593 for the salvage attempt (also closed).

PR #3445, opened 2026-07-08
[Godfile data-store step 1: extract sqlite-migrations](1)

EdbertChan:
## Summary

Slice 1 of the data-store god-file decomposition.

This moves the SQLite migration routines out of the 3,906-line sqlite-adapter.ts into a new sqlite-migrations.ts module.

The adapter keeps the same public methods; each one now delegates to a free function over a narrow SqliteExecutor context.

A pure sqlite-external-dependencies.ts helper is also split out so a later repository slice can reuse it.

The moved code used to hide errors.

Several catch blocks discarded every caught error behind a "table may not exist yet" comment, which cannot happen because the tables are created before migrations run.

Those blocks now log the caught error with its function name and row context, so a real database or JSON failure is debuggable instead of silently vanishing.

Migrations still tolerate missing tables and never abort startup. The one added behavior is that caught errors now warn instead of vanishing.

The workflow-dependency migration also stops dropping data: it now clears a task's external_dependencies only when those deps were merged into an existing workflow.

## Review Claim

The migrate* logic moves from sqlite-adapter.ts to sqlite-migrations.ts as free functions over a SqliteExecutor context, the adapter methods become one-line delegates, every caught migration error is now logged with context instead of being silently swallowed, and the workflow-dependency migration clears only the tasks whose deps were promoted to an existing workflow.

## Review Lane

behavior

## Review Unit

ownership-refactor

## Safety Invariant

Migration order and control flow are unchanged: migrations run from the same call sites, still tolerate missing tables and malformed rows without aborting, and the COLUMN_MIGRATIONS loop still rethrows anything but the idempotent "duplicate column name". Two intended behavior changes: caught errors now log through `console.warn`, and `migrateTaskExternalDependenciesToWorkflows` clears `external_dependencies` only for tasks whose deps were promoted to an existing workflow, so malformed or orphaned rows are no longer silently dropped. Both are covered by `sqlite-migrations-error-logging.test.ts`.

## Slice Rationale

One extraction per slice, following the review-compression convention. The silent-swallow catch blocks were introduced into the new module by this very move, so repairing them belongs in the same slice rather than a follow-up; the change stays inside the extracted file plus a focused test for it.

## Non-goals

- No task/attempt or workflow repository extraction; those are later slices.
- No change to PersistenceAdapter (adapter.ts), the package barrel index.ts, or any SQL text.
- No new migrations and no reordering of migration execution.
- No change to migration order or the public SQLiteAdapter method surface.

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `pnpm --filter @invoker/data-store exec vitest run src/__tests__/sqlite-migrations-error-logging.test.ts src/__tests__/migrate-gate-policy.test.ts src/__tests__/sqlite-adapter.test.ts`
- [ ] `pnpm --filter @invoker/data-store exec vitest run` (full data-store suite: 362 passed)
- [ ] `pnpm run check:types`

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

</details>

### DIFF
--- a/packages/data-store/src/__tests__/sqlite-migrations-error-logging.test.ts
+++ b/packages/data-store/src/__tests__/sqlite-migrations-error-logging.test.ts
@@ -1,0 +1,110 @@
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import * as migrations from '../sqlite-migrations.js';
+import type { SqliteExecutor } from '../sqlite-executor.js';
+
+/**
+ * Guards the CLAUDE.md invariant for this module: migrations tolerate missing
+ * tables and malformed rows so a partial upgrade never aborts startup, but a
+ * caught error is logged with context — never silently swallowed.
+ */
+function stubExecutor(overrides: Partial<SqliteExecutor> = {}): SqliteExecutor {
+  const notExist = (): never => {
+    throw new Error('no such table: tasks');
+  };
+  return {
+    queryOne: notExist,
+    queryAll: notExist,
+    execRun: () => {},
+    runTransaction: <T>(work: () => T): T => work(),
+    run: () => {},
+    getRowsModified: () => 0,
+    readOnly: false,
+    markDirty: () => {},
+    ...overrides,
+  };
+}
+
+describe('sqlite-migrations error logging', () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('surfaces the caught error instead of swallowing when a query fails', () => {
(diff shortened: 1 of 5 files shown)


EdbertChan:

3445. #3445 — [Godfile data-store step 1: extract sqlite-migrations](1) ← this PR
   Base: master
   Link: https://github.com/Neko-Catpital-Labs/Invoker/pull/3445

PR #3455, opened 2026-07-08
fix(workflow-core): cancelled never-started dependents become blocked, not failed

EdbertChan:
## Summary

When a running task is cancelled, Invoker used to mark every task waiting behind it as failed — even tasks that never started.

That red badge lied: it implied a run that never happened.

Now a never-started dependent of a cancelled task becomes blocked, with a short reason. Only the task you actually stopped stays failed.

A blocked dependent frees itself once the upstream finishes, and the workflow still reports failed, so nothing merges by mistake.

A smaller companion change keeps a workflow from looking active when its only leftover work is a dependent stuck behind a cancelled task.

That prevents an in-place graph edit from becoming an accidental fork.

## Review Claim

Approve turning a cancelled task's never-started dependents into blocked (not failed), and treating such stuck dependents as not-live so in-place replace is preserved.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

Pure orchestrator logic, covered by workflow-core unit checks (840 pass) and a full typecheck. The dependency readiness gate is untouched, so ordering still holds and nothing can run before its upstream finishes; only the terminal-status labelling on cancel and the workflow-live predicate change. The task actually cancelled still ends failed, and `failed` outranks `blocked` in the rollup, so the workflow stays failed and the merge gate stays closed.

## Slice Rationale

This is one behavior fix plus its own regression test, kept as a single slice because it changes exactly one thing: what status a never-started dependent gets when its upstream is cancelled.

## Non-goals

- Does not change the status of the task you directly cancel; that task stays failed.
- Does not touch the dependency scheduler readiness gate, so a task still cannot run before its upstream finishes.
- Does not change the existing cross-workflow external-gate blocking path (those blocked tasks stay live).
- Does not add UI code; it reuses the existing `blocked` badge already rendered today.
- Does not add a changelog or documentation entry (kept out to preserve behavior-lane atomicity).

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `pnpm --filter @invoker/workflow-core test`
- [ ] `pnpm run check:types`

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

</details>

### DIFF
--- a/packages/workflow-core/src/__tests__/cancel-task.test.ts
+++ b/packages/workflow-core/src/__tests__/cancel-task.test.ts
@@ -175,17 +175,33 @@ describe('cancelTask', () => {
     expect(task!.execution.error).toContain('Terminated by user');
   });
 
-  it('cascades cancel to pending dependents', () => {
+  it('cascades cancel to never-started dependents as blocked, not failed', () => {
     orchestrator.loadPlan(simplePlan());
 
     orchestrator.cancelTask('a');
 
+    const taskA = orchestrator.getTask('a');
     const taskB = orchestrator.getTask('b');
     const taskC = orchestrator.getTask('c');
-    expect(taskB!.status).toBe('failed');
-    expect(taskC!.status).toBe('failed');
-    expect(taskB!.execution.error).toContain('upstream task "a"');
-    expect(taskC!.execution.error).toContain('upstream task "a"');
+
+    // The cancel root stays failed — it is what the operator terminated.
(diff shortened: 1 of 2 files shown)


EdbertChan:
Superseded by the [Blocked Cascade] stack: #3473 (this same fix, now the stack base) and #3474 (follow-up hardening of blocked-task liveness). Closing in favor of the stack.

PR #3698, opened 2026-07-09
Orchestrator self-heals stale selected_attempt_id

EdbertChan:
## Summary

A workflow card kept rendering a red "Failed: see details" pill for a task whose underlying DB row had no error.

The task's `selected_attempt_id` still pointed at an older `failed` attempt even though a strictly newer active attempt already existed.

`SQLiteAdapter.reconcileTaskFromSelectedAttempt` reads status and error off the selected attempt when the task itself is non-terminal, so the outdated pointer leaked the old failure to the UI.

This PR routes a self-healing pass through `syncAllFromDb` / `syncFromDb`, the existing hydration lifecycle events.

If the selected attempt is failed or superseded and a newer active attempt exists, the pointer advances and the injected terminal fields are cleared.

Reconciliation logs a `task.selected_attempt_advanced` event so the swap is auditable in the activity log.

## Review Claim

`Orchestrator.reconcileStaleSelectedAttempts()`, routed through `syncAllFromDb` / `syncFromDb`, is the right place to heal `selected_attempt_id` drift so the projection stops emitting an outdated `failed` view of a task that has a newer active attempt.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

The reconciliation only touches a task when its currently selected attempt is `failed` or `superseded` AND a strictly newer attempt with `isActiveAttempt(...)` is present. A `completed` selected attempt is intentionally left alone — that is a legitimate hand-off state used by merge and fix flows. The pointer never regresses in time because the newer candidate must have `createdAt > selected.createdAt`.

The write goes through `writeAndSync(..., { skipWorkflowStatusSync: true })` so the sweep does not spuriously toggle workflow rollups; the caller's own status sync in `syncAllFromDb` / `syncFromDb` still runs afterwards via `assertMergeLeavesInvariant`.

## Slice Rationale

The bug reproduced against real production data — a cancellation-then-fresh-attempt sequence that left the pointer outdated. Fixing it in `SQLiteAdapter.reconcileTaskFromSelectedAttempt` would only hide the failed row on the read path — the orchestrator would keep leaking an outdated pointer, and other callers of `loadTasks` would still see the failed projection. Healing the pointer at hydration is the root-cause fix and keeps the DB self-consistent for every reader.

## Non-goals

- Not changing how `selected_attempt_id` is set during normal task lifecycle transitions.
- Not changing `SQLiteAdapter.reconcileTaskFromSelectedAttempt`; the projection stays as-is for tasks whose pointer is legitimately terminal.
- Not adding a background job or scheduler; the sweep piggybacks on the existing hydration entry points.
- Not touching merge-node or gate reconciliation invariants.

### Before

```mermaid
graph TD
    A["syncAllFromDb / syncFromDb"] --> B["stateMachine.restoreTask(...)"]
    B --> J["assertMergeLeavesInvariant"]
    B -.projection.-> P["SQLiteAdapter.reconcileTaskFromSelectedAttempt"]
    P -.reads.-> Q["selected_attempt_id -> failed attempt"]
    Q -.emits.-> R["task.status = failed, task.execution.error set"]
    R -.consumed by.-> S["buildActionGraphDiagnostics -> blocker:{taskId}:error"]
    S -.renders.-> T["Failed: see details pill"]
```

### After

```mermaid
graph TD
    A["syncAllFromDb / syncFromDb"] --> B["stateMachine.restoreTask(...)"]
    B --> C["reconcileStaleSelectedAttempts()"]
    C --> D{"selected.status in {failed, superseded}?"}
    D -- "no" --> E["skip"]
    D -- "yes" --> F["loadAttempts(task.id)"]
    F --> G{"newer active attempt exists?"}
    G -- "no" --> E
    G -- "yes" --> H["writeAndSync: advance pointer, clear terminal fields"]
    H --> I["logEvent task.selected_attempt_advanced"]
    C --> J["assertMergeLeavesInvariant"]
```

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `cd packages/app && pnpm test src/__tests__/action-graph-snapshot.test.ts`

</details>

## Visual Proof

The user-visible effect is that a workflow card in the DAG stops showing the red "Failed: see details" pill once a newer active attempt exists. A static screenshot alone cannot demonstrate this behavior because the buggy state is only reachable by a specific timing sequence — workflow cancellation while a leased launch dispatch is in flight, followed by a fresh attempt — that cannot be dropped into a demo screen deterministically.

The inspectable state proof lives in `packages/app/src/__tests__/action-graph-snapshot.test.ts`, which constructs exactly the stuck state observed in production, runs `buildCurrentActionGraphSnapshot`, and asserts the failed blocker is absent from the graph and the `task.selected_attempt_advanced` audit event is present. That is the same signal the UI reads to render the pill.

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None. The revert restores the prior behavior where outdated `selected_attempt_id` pointers surface as "Failed: see details" pills. Existing DB rows are untouched by this change.
- Data migration? No. The reconciliation runs at hydration and writes only if drift is detected; no schema or migration changes.

</details>

### DIFF
--- a/packages/app/src/__tests__/action-graph-snapshot.test.ts
+++ b/packages/app/src/__tests__/action-graph-snapshot.test.ts
@@ -1,0 +1,100 @@
+import { afterEach, describe, expect, it } from 'vitest';
+import { SQLiteAdapter } from '@invoker/data-store';
+import { InMemoryBus } from '@invoker/test-kit';
+import { Orchestrator, type Attempt, type TaskState } from '@invoker/workflow-core';
+import { buildCurrentActionGraphSnapshot } from '../action-graph-snapshot.js';
+import type { InvokerConfig } from '../config.js';
+
+describe('buildCurrentActionGraphSnapshot', () => {
+  const adapters: SQLiteAdapter[] = [];
+
+  afterEach(() => {
+    for (const adapter of adapters.splice(0)) {
+      adapter.close();
+    }
+  });
+
+  it('does not emit a failed blocker when the selected_attempt_id is stale but a newer active attempt exists', async () => {
+    const adapter = await SQLiteAdapter.create(':memory:');
+    adapters.push(adapter);
+
+    const workflowId = 'wf-stale';
+    const taskId = 'wf-stale/verify';
+    adapter.saveWorkflow({
+      id: workflowId,
+      name: 'Remove Running Sidebar And Workers Screen Title',
+      createdAt: new Date().toISOString(),
+      updatedAt: new Date().toISOString(),
+    });
+
+    const olderCreatedAt = new Date('2026-07-09T05:38:02.000Z');
+    const newerCreatedAt = new Date('2026-07-09T05:51:12.808Z');
+
+    const failed: Attempt = {
+      id: `${taskId}-aOLD`,
+      nodeId: taskId,
+      queuePriority: 0,
+      status: 'failed',
+      upstreamAttemptIds: [],
+      error: 'Cancelled by user (workflow)',
+      completedAt: new Date('2026-07-09T05:51:12.761Z'),
+      startedAt: new Date('2026-07-09T05:43:54.107Z'),
+      lastHeartbeatAt: new Date('2026-07-09T05:43:54.107Z'),
+      createdAt: olderCreatedAt,
+    };
+    const newerPending: Attempt = {
+      id: `${taskId}-aNEW`,
+      nodeId: taskId,
+      queuePriority: 0,
+      status: 'pending',
+      upstreamAttemptIds: [],
+      supersedesAttemptId: failed.id,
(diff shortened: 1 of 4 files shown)


EdbertChan:
Pivoting to a read-side projection fix per review feedback: 
- Self-heal on every hydration is a symptom fix; the projection layer is where the outdated view is amplified.
- Follow-up PR will make `SQLiteAdapter.reconcileTaskFromSelectedAttempt` prefer the newest active attempt when the pointer targets a discarded/failed attempt but a newer active attempt exists.
- Separate write-side investigation issue will track down which code path created the newer attempt without advancing `selected_attempt_id` in the first place.

PR #3706, opened 2026-07-09
fix(headless): answer delegated `worker status` on the running owner

EdbertChan:
## Summary

The `worker status` headless command failed whenever the desktop app was running, so operators could not check auto-fix worker health from the CLI.

The running app owns the database. A read-only command from a second process is answered by that owner.

The owner had no handler for `worker`, so the delegated command errored instead of returning status.

Now the owner answers `worker status` and returns the diagnostic, matching the standalone command.

## Review Claim

The running owner answers a delegated `worker status` headless command instead of erroring, and mutating `worker <kind>` commands stay non-delegatable.

## Review Lane

behavior

## Review Unit

activation-surface

## Safety Invariant

The shared renderer only reads worker audit rows through the existing persistence accessor and writes through the same output sink the query path already uses. The owner still refuses to delegate mutating `worker <kind>` scans, so no new mutation path is opened; classification is unchanged.

## Slice Rationale

One focused headless command fix plus its regression test; it changes only the `worker` command handling and shares a single status renderer between the standalone and owner code paths.

## Non-goals

- Does not change the mutating `worker <kind>` scan path, which still runs standalone and acquires the worker lock.
- Does not make `worker list` delegatable; only `worker status` is answered by the owner.
- Does not change the auto-fix worker's behavior, retry budget, or scheduling.
- Does not touch command classification; `worker` stays read-only in the registry.

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `pnpm --filter @invoker/app exec vitest run src/__tests__/headless-query-capture.test.ts`
- [ ] `pnpm run check:types`

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

</details>

### DIFF
--- a/packages/app/src/__tests__/headless-query-capture.test.ts
+++ b/packages/app/src/__tests__/headless-query-capture.test.ts
@@ -112,4 +112,30 @@ describe('runReadOnlyHeadlessQueryToString', () => {
     await expect(runReadOnlyHeadlessQueryToString(['query', 'ui-perf'], deps)).resolves.toContain('mainDeltaToUi');
     expect(resetUiPerfStats).not.toHaveBeenCalled();
   });
+
+  it('answers a delegated `worker status` query on the writable owner', async () => {
+    const deps = makeQueryDeps(() => []);
+    deps.persistence = {
+      listWorkflows: () => [],
+      loadTasks: () => [],
+      getEvents: () => [],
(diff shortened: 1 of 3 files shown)


EdbertChan:
Superseded by #3711, which republishes the same change (3 files, +88/-46) as a Mergify stack so it can be landed via the land-stack guard. No code differences.

PR #4114, opened 2026-07-11
Planning Lag Step 1 Telemetry Baseline

EdbertChan:
## Summary

Invoker's planning view feels laggy when many chats are open. Before we can fix it, we need a way to measure that lag.

This adds a passive, invisible telemetry probe. When you type in any editable field, it records how long the edit took and how much planning state was loaded at that moment (chat count, task count, transcript size).

Nothing on screen changes. The probe only emits a metric through the existing `reportUiPerf` channel, and only when that channel exists.

A regression test locks in a deterministic baseline for a busy many-chats workspace, so future changes can be compared against a known reference.

## Review Claim

Approve adding a passive `planning_typing_lag` telemetry probe plus its deterministic baseline test, with no change to any rendered UI.

## Review Lane

`behavior`

## Review Unit

`planning-typing-telemetry`

## Safety Invariant

The probe is safe to review locally because it cannot alter what the user sees or blocks. It attaches one capture-phase, passive `input` listener that early-returns unless `window.invoker.reportUiPerf` exists, skips read-only, disabled, and composing targets, throttles to one report per 250ms, defers the emit to `requestAnimationFrame`, guards for absent `window`/`document` (SSR), and removes the listener plus cancels any pending frame on unmount.

## Slice Rationale

The instrumentation and its baseline test are one review unit: the test asserts the exact metric shape the probe emits, so neither half is meaningful alone and splitting them would land a red slice. The workflow's separate verify task produced no file changes (an empty test-run commit), so it is folded here rather than published as its own no-op PR.

## Non-goals

- Does not change any rendered UI, layout, or user-visible text.
- Does not attempt to reduce the typing lag itself; this is measurement only.
- Does not add dashboards, alerts, or server-side aggregation.
- Does not modify the `reportUiPerf` transport or its consumers.

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `cd packages/ui && pnpm install --frozen-lockfile`
- [ ] `cd packages/ui && pnpm exec vitest run src/__tests__/invoker-terminal.test.tsx --testTimeout=30000`

Both commands were run; the telemetry file passes 2/2 tests. The `--testTimeout=30000` is required because the many-chats baseline test takes ~5.5s to render and exceeds vitest's default 5s timeout under load.

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None. Reverting removes the listener and the test; no other code depends on the `planning_typing_lag` metric.
- Data migration? No

</details>

### DIFF
--- a/packages/ui/src/App.tsx
+++ b/packages/ui/src/App.tsx
@@ -11,7 +11,7 @@
 
 import { useState, useCallback, useMemo, useEffect, useRef, useLayoutEffect } from 'react';
 import yaml from 'js-yaml';
-import type { TaskState, TaskReplacementDef, ExternalGatePolicyUpdate, WorkflowStatus } from './types.js';
+import type { TaskState, TaskReplacementDef, ExternalGatePolicyUpdate, WorkflowMeta, WorkflowStatus } from './types.js';
 import type { ActionGraphNode } from '@invoker/contracts';
 import { useTasks } from './hooks/useTasks.js';
 import { useInvoker } from './hooks/useInvoker.js';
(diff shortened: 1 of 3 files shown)


EdbertChan:
### Revision history

| 1 | initial | `424d6d1` |  | 2026-07-11 08:54 UTC |
| 2 | rebase | `424d6d1 → 5ca621a` _(rebase only)_ |  | 2026-07-11 08:54 UTC |
| 3 | rebase | `5ca621a → 424d6d1` _(rebase only)_ |  | 2026-07-11 08:55 UTC |

EdbertChan:
Superseded by the review-compressed 2-PR stack #4116 (instrument, `behavior`) → #4117 (lock baseline with regression test, `proof`). This single PR bundled the probe and its regression test into one slice, which mixes behavior + proof and trips the diff-atomicity / review-compression rule. The split stack has identical combined content (same final tree) and is the canonical publication for this work.

PR #4618, opened 2026-07-15
[Owner Rail Cleanup](1) Drop legacy Run and Stop controls

EdbertChan:
## Summary

Start ready work is the owner activation entry point now.

The legacy Run and Stop rail buttons targeted an older start/stop lifecycle that no longer matches fire-and-forget ready work.

This slice drops those controls and updates planner copy to reference Start ready work.

## Review Claim

Approve dropping legacy Run and Stop controls so the owner rail exposes only Start ready work.

## Review Lane

behavior

## Review Unit

activation-surface

## Safety Invariant

Start ready work behavior is unchanged.

This slice only drops obsolete controls and updates user-facing planner text.

## Slice Rationale

Graph left-click selection is a separate UI regression fixed in the prior slice.

Rail control changes are isolated so reviewers can approve them without mixed graph changes.

## Non-goals

- No Start ready work handler changes
- No workflow graph interaction changes
- No worker stop controls in the Workers panel

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `cd packages/ui && pnpm exec vitest run src/__tests__/invoker-terminal.test.tsx`

</details>

## Visual Proof

Owner rail shows Start ready work without legacy Run or Stop buttons.

Owner rail with Start ready work only

## Revert Plan

<details>
<summary>Revert Plan</summary>

- Safe to revert? Yes
- Revert command: `git revert <sha>`
- Post-revert steps: None
- Data migration? No

</details>

## Summary by CodeRabbit

* **New Features**
  * Replaced the legacy Run/Stop workflow with a unified **Start ready work** action.
  * Added readiness checks before starting work and clearer task-specific success or failure messages.
  * Updated the interface to support previewing or recreating failed ready work.

* **Documentation**
  * Updated planning guidance and tutorial text to reference **Start ready work** instead of “Run it.”

PR #4724, opened 2026-07-16
Add a master switch to turn all workers on and off

EdbertChan:
## Summary

The Workers panel only had an Enable/Disable button on each worker row. To stop everything you had to click every worker one at a time.

This adds one master switch above the worker rows: **Turn workers off** / **Turn workers on**. It appears on the Workers page and in the Queue view.

Turning it off stops every running worker and keeps them off across restarts. Turning it on brings back exactly the workers that were on before.

The switch is stored separately from each worker's own setting. A worker you deliberately disabled stays disabled after an off/on cycle.

It defaults to on, so existing installs behave exactly as they do today.

## Review Claim

One master switch turns all workers off and back on, without overwriting any individual worker's own enabled setting.

## Review Lane

behavior

## Review Unit

activation-surface

## Safety Invariant

The global flag lives in its own `worker_global_state` row and is read-only to the per-kind path. `setGlobalEnabled` never calls `setWorkerDesiredState`, so per-worker intent survives the cycle.

Absent rows read as enabled, so existing databases keep their current auto-start behavior with no migration step.

## Slice Rationale

The switch is one user-facing control. Its persistence, IPC channel, engine gating, and UI are the minimum that makes it work; a smaller slice would ship a switch wired to nothing.

## Non-goals

Does not change what any individual worker does, or its polling interval.

Does not restore the rail Run/Stop task-execution controls removed in #4618. Those drive `invoker:start` / `invoker:stop`, not workers.

Does not add the switch to the read-only web surface, which renders worker controls disabled already.

## Architecture

Worker start-up now passes through two gates instead of one. The master switch gates the per-worker setting; it does not overwrite it.

### Before

```mermaid
graph TD
    A["startAutoStartedWorkers()"] --> B["desiredEnabledForKind(kind)"]
    B -->|true| C["runtime.start()"]
    B -->|false| D["stays stopped"]
```

### After

```mermaid
graph TD
    A["startAutoStartedWorkers()"] --> G{"globalEnabled?"}
    G -->|false| H["no worker starts"]
    G -->|true| B["desiredEnabledForKind(kind)"]
    B -->|true| C["runtime.start()"]
    B -->|false| D["stays stopped"]
    S["setGlobalEnabled(false)"] --> T["stopAll() — desired state untouched"]
```

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `cd packages/app && pnpm test` — 22 pass in `worker-control.test.ts`, incl. off/on cycle preserving per-worker state
- [ ] `cd packages/ui && pnpm test` — 773 pass, incl. 8 in `worker-global-switch.test.tsx`
- [ ] `cd packages/data-store && pnpm test` — 398 pass, incl. global-switch round-trip and default-on
- [ ] `cd packages/contracts && pnpm test` — 87 pass
- [ ] `cd packages/app && CAPTURE_MODE=after npx playwright test visual-proof.spec.ts -g "global worker switch"` — 2 pass; drives the real switch in Electron from both the Workers surface and the Queue view
- [ ] `pnpm build` — clean across all packages
- [ ] `pnpm run check:comments && pnpm run check:types` — clean

Not fixed here, fails identically on unmodified `master`: `focus-switch-main-process-cost.test.ts` perf budget (~4.8s vs 100ms) on this machine.

</details>

## Visual Proof

Captured by `visual-proof.spec.ts`, driving the real switch in Electron on both surfaces that show workers.

### Workers surface (sidebar → Workers)

**Walkthrough (on → off → on):**

Master switch on the Workers surface

**Step 1 — on.** Header reads "Turn workers off". Requeue shows a green `Process: Running`; the sidebar reads "9 processes running".

Step 1: workers on

**Step 2 — off.** Header reads "Turn workers on". Requeue flipped to `Process: Stopped` and the sidebar reads "0 processes running". Requeue keeps its `Starts on launch` chip — that is the preserved per-worker setting.

Step 2: workers off

**Step 3 — back on.** Requeue returns to `Process: Running` and the sidebar returns to "9 processes running". Autofix was off before the cycle and is still `Process: Stopped` — the switch did not turn it on.

Step 3: back on, previously-off workers still off

### Queue view

Same switch, above the per-worker rows. Here the per-worker "Enable worker" buttons grey out while the switch is off.

Master switch in the Queue view

## Revert Plan

<details>

- Safe to revert? Yes
- Revert command: `git revert cbf01f18b 7a6a6cf35 d8ef2a7ff`
- Post-revert steps: None. Workers return to auto-starting from per-kind desired state.
- Data migration? No. The `worker_global_state` table is additive and simply goes unread; a stale `enabled = 0` row is ignored after revert, so workers resume.

</details>

### DIFF
--- a/packages/app/e2e/visual-proof.spec.ts
+++ b/packages/app/e2e/visual-proof.spec.ts
@@ -2139,6 +2139,31 @@ test.describe('Visual proof capture', () => {
 
 
 
+  test('global worker switch turns all workers off and back on', async ({ page }) => {
+    await loadPlan(page, TEST_PLAN);
+    await selectGraphMenuItem(page, 'rail-queue');
+    await expectQueueViewVisible(page);
+
+    const toggle = page.getByTestId('worker-global-switch');
+    await expect(toggle).toBeVisible();
+    await expect(toggle).toHaveAttribute('aria-checked', 'true');
+    await expect(toggle).toHaveText('Turn workers off');
+    await expect(page.getByTestId('worker-process-list')).toContainText('Running');
+    await captureScreenshot(page, 'worker-global-switch-step-1-on');
+
+    await toggle.click();
+    await expect(toggle).toHaveAttribute('aria-checked', 'false');
+    await expect(toggle).toHaveText('Turn workers on');
+    await expect(page.getByTestId('worker-process-list')).not.toContainText('Running');
+    await expect(page.getByTestId('worker-start-stop-pr-status')).toBeDisabled();
+    await captureScreenshot(page, 'worker-global-switch-step-2-off');
+
+    await toggle.click();
+    await expect(toggle).toHaveAttribute('aria-checked', 'true');
+    await expect(page.getByTestId('worker-process-list')).toContainText('Running');
+    await captureScreenshot(page, 'worker-global-switch-step-3-back-on');
+  });
+
   test('queue view concurrency display', async ({ page }) => {
     await loadPlan(page, TEST_PLAN);
     const now = new Date();
--- a/packages/app/e2e/visual-proof.spec.ts
+++ b/packages/app/e2e/visual-proof.spec.ts
@@ -2139,6 +2139,28 @@ test.describe('Visual proof capture', () => {
 
 
 
+  test('global worker switch turns all workers off and back on from the Workers surface', async ({ page }) => {
+    await loadPlan(page, TEST_PLAN);
+    await page.getByTestId('sidebar-workers').click();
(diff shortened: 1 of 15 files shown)


EdbertChan:
Closing — wrong design. This built one master switch for all workers; what was wanted is an off switch per worker, on the Workers page.

Replaced by #4732, which gives each worker row on the Workers page its own Enable/Disable button. No global switch, no schema change, no new IPC channel.

The branch `feat/global-worker-switch` is abandoned; nothing from it was kept.

PR #4749, opened 2026-07-16
Reclaim pr-cron-work and stale /tmp scratch on critical disk pressure

EdbertChan:
## Summary

A droplet in the SSH pool filled to 100% disk and stayed there. Work pinned to it waited forever while the disk-headroom reclaim reported success on every pass.

The reclaim only freed a fixed set of dirs under `~/.invoker`. It never touched `pr-cron-work` (3.3G of leftover PR clones) or `/tmp`.

But `/tmp` is where the test scratch that overflows the box lives.

This adds `pr-cron-work` to the reclaimable dirs and teaches the remote reclaim script to free Invoker and test scratch from the shared temp dir.

The `/tmp` pass is deliberately narrow. It matches only known scratch name patterns, plus stray `mktemp` leftovers older than 60 minutes.

It excludes system dirs, sockets, and lockfiles, and never wipes the whole temp dir, so an in-flight run keeps files it is still writing.

It also preserves any temp entry holding a mineable `.jsonl` agent-session transcript, so session data we can mine survives.

## Review Claim

The critical-disk reclaim now frees `pr-cron-work` and old `/tmp` scratch, and never wipes `/tmp` wholesale.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

The `/tmp` pass matches an explicit allowlist of scratch name patterns and only frees `mktemp` leftovers older than 60 minutes, excluding `systemd-private-*`, `snap-*`, `.*-unix`, `ssh-*`, `claude-*`, and `*.lock`. There is no `rm -rf /tmp` and no wipe of the whole temp dir; the tests assert both are absent. Every removal routes through a guard that preserves any entry holding a `.jsonl`, so mineable agent-session transcripts survive.

## Slice Rationale

One cohesive change to the reclaim target set. Both new targets are regenerable scratch freed under the same critical-pressure trigger and the same guards, so they belong in one slice.

## Non-goals

Does not change the disk-pressure thresholds, the cooldown, or which pool members are probed. Does not change how launches are routed or dispatched. Does not touch the separate 7-day `pr-cron-work` prune in the PR-maintenance cron.

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `cd packages/execution-engine && pnpm exec vitest run src/__tests__/disk-headroom-reclaim.test.ts`
[...]

PR #4750, opened 2026-07-16
[Disk reclaim orphans](2) Reap stale /tmp scratch homes, local + remote

EdbertChan:
## Summary

A remote worker box filled its disk to 100 percent. A full disk makes every git clone on that box fail.

Tasks then ended before a workspace folder was ever written, which is what surfaced the confusing state to the operator.

The disk-headroom worker did notice the full disk. It just could not free enough room.

Today it only empties a few fixed folders inside a known Invoker home.

It never reaps the orphaned `/tmp/invoker-*` end-to-end scratch folders, which had grown to roughly 50 GB on that box.

This change teaches the same worker to reclaim that leftover temp scratch on both the local and remote paths.

An age guard keeps any folder touched in the last hour, so a live run is never disturbed.

## Review Claim

The disk-headroom worker now frees leftover Invoker temp scratch on both the local and remote paths, while keeping recent folders, the live Invoker home, and system or lock folders safe.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

Removal is bounded three ways: entries younger than 60 minutes are left alone (a live run may still hold them), the live Invoker home is excluded from every sweep, and system and lock names (`systemd-private-*`, `ssh-*`, `claude-*`, `*.lock`, and similar) are always preserved. The existing path guards that refuse `/`, `~`, and `$HOME` are unchanged. Unit tests assert both the deletions and the preservations for the local path, and assert the age gate and home exclusion for the remote script.

## Slice Rationale

This is one focused behavior change to the disk-headroom worker. It finishes an in-tree partial edit that had only touched the remote path, and closes an age-guard hole in that edit, so both belong together as one reviewable idea.

## Non-goals

Does not change disk detection, thresholds, cooldown, or the worker schedule. Does not change the set of reclaimable Invoker-home subfolders. Does not touch the "no saved workspace" message or the fix-with-agent gate that surfaces it.

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `cd packages/execution-engine && npx vitest run src/__tests__/disk-headroom-reclaim.test.ts`
- [ ] `cd packages/execution-engine && pnpm build`

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Revert command: `git revert 7b0d0291f`
- Post-revert steps: None. Reverting restores the prior behavior, where the worker does not reap temp scratch. No data is deleted or migrated by the revert.
- Data migration? No

</details>

### DIFF
--- a/packages/execution-engine/src/__tests__/disk-headroom-reclaim.test.ts
+++ b/packages/execution-engine/src/__tests__/disk-headroom-reclaim.test.ts
@@ -1,4 +1,12 @@
-import { mkdtempSync, mkdirSync, writeFileSync, existsSync, readdirSync, rmSync } from 'node:fs';
+import {
+  mkdtempSync,
+  mkdirSync,
+  writeFileSync,
+  existsSync,
+  readdirSync,
+  rmSync,
+  utimesSync,
+} from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { afterEach, describe, expect, it, vi } from 'vitest';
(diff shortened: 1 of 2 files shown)


EdbertChan:
Superseded. The /tmp scratch reaping this adds already landed via #4749; #4750 duplicated it and conflicts. Its one real safety improvement — age-gating the targeted glob reap so it can't delete a running task's fresh scratch, plus an $INVOKER_HOME guard — is salvaged in #4761. Closing in favor of that.

PR #4846, opened 2026-07-17
Fix pr-ci-failure-scan worker: load shared lib from repo root, not sibling dir

EdbertChan:
## Summary

One execution-engine maintenance worker failed on every scheduled tick and never got past its own startup. This restores it.

The worker runtime routes each scheduled tick to a shell entrypoint. That entrypoint could not load the shared library it depends on, so it aborted before doing any work.

The entrypoint lives in a package subfolder but loaded the shared library from its own directory, where the library is absent. The library lives in the repo-root scripts directory.

The fix resolves the shared library from the repo root, so where the entrypoint lives no longer matters. Two sibling entrypoints in the same routing layer already load it correctly.

This does not change what the pr-ci-failure-scan worker does once it runs, and it does not touch the shared library itself.

A startup guard now drives every routed maintenance entrypoint with its lock held, so each one must load its shared library or the guard fails.

## Review Claim

The routed maintenance entrypoint now loads its shared library and can start, and a startup guard keeps every entrypoint in the worker routing layer from regressing the same way.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

The change only alters where the entrypoint resolves `cron-pr-lib.sh` (repo root instead of a non-existent sibling path). The new guard spawns each real entrypoint with a `flock` stub that forces the lock-held early exit, so it proves the source line resolves without touching gh, git, or the network.

## Slice Rationale

The path fix and the guard that proves it are one routing change: the fix restores how the worker runtime reaches this entrypoint, and the guard fails without it. They are one claim, not two.

## Non-goals

Does not relocate the entrypoint into `scripts/` alongside its siblings, does not change the scan's actual PR logic, and does not touch the shared `cron-pr-lib.sh` library.

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `cd packages/execution-engine && npx vitest run src/__tests__/pr-maintenance-entrypoint-bootstrap.test.ts`
- [ ] With a `flock` stub that exits 1 on `PATH`: `INVOKER_PR_CRON_LOCK=/tmp/stub/lock bash packages/execution-engine/scripts/cron-pr-ci-failure.sh` — expect exit 0 and no "No such file or directory"

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None. Reverting restores the prior broken load line; the worker returns to exiting 1 each tick.
- Data migration? No

</details>

### DIFF
--- a/packages/execution-engine/scripts/cron-pr-ci-failure.sh
+++ b/packages/execution-engine/scripts/cron-pr-ci-failure.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-# shellcheck source=scripts/cron-pr-lib.sh
-source "$(dirname "$0")/cron-pr-lib.sh"
+# shellcheck source=../../../scripts/cron-pr-lib.sh
+source "$(cd "$(dirname "$0")/../../.." && pwd)/scripts/cron-pr-lib.sh"
 
 cron_lock
 
--- a/packages/execution-engine/src/__tests__/pr-maintenance-entrypoint-bootstrap.test.ts
+++ b/packages/execution-engine/src/__tests__/pr-maintenance-entrypoint-bootstrap.test.ts
@@ -1,0 +1,65 @@
+import { spawnSync } from 'node:child_process';
+import { chmodSync, existsSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join, resolve } from 'node:path';
+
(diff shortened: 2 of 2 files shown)


EdbertChan:
### Revision history

| 1 | initial | `49ecd61` |  | 2026-07-17 04:27 UTC |
| 2 | content | [`49ecd61 → efd2347`](https://github.com/Neko-Catpital-Labs/Invoker/compare/49ecd610a9b51d55fe2ab37e8ee802ee6f5d1dd6...efd2347b1582b22b539136439b878b57c9fc4d4c) |  | 2026-07-17 04:27 UTC |

PR #5810, opened 2026-07-25
[Owner Build Guard] Refuse to delegate to a build-mismatched owner

EdbertChan:
## Summary

Invoker keeps one "owner" process holding the database; every other process forwards reads and writes to it over the message bus.

That owner's build was never checked. A process running different, older code could hold the owner role indefinitely while a newer process still treated it as fully compatible.

Once treated as compatible, every request the old owner's code didn't recognize failed on its own, one at a time, instead of one clear early failure.

The owner-ping handshake now carries each process's build identity — the git commit it was built from.

A process that finds a mismatched, or identity-less pre-fix, owner won't treat it as capable. It fails once, with one readable reason.

## Review Claim

The owner lookup should refuse to treat a build-mismatched owner as capable, instead of always trusting whatever process answers the ping.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

The new field is optional and only present on a mismatch, so the existing shape returned to callers is unchanged in the common case.

The comparison always no-ops when the local process isn't a real build (every test run, and any non-packaged run), so no existing assertion needed a change for this to land safely.

## Slice Rationale

This is a self-contained fix for one observed incident (an old, globally-installed Invoker build silently answering a newer process's message-bus requests) and is independent of the unrelated PR-babysit work in flight on another branch.

## Non-goals

Does not change what a user sees on screen — a mismatch is currently only written to the process log and to stderr at the two ownership call sites in `main.ts`; a friendlier in-app notice is a follow-up if wanted.

Does not add a hand-maintained protocol number — it compares the exact build (git commit) instead, which needs no one to remember to bump a counter.

Does not touch the separate, already in-flight fix for a dead process's leftover lock file (a different failure mode from a live-but-mismatched owner).

### Before

```mermaid
graph TD
    A["process pings the owner"] --> B["owner answers with ownerId/mode"]
    B --> C["capable = mode is a known one"]
    C --> D["every request is forwarded to the owner"]
    D --> E["owner doesn't recognize the request"]
    E --> F["one failure per request"]
```

### After

```mermaid
graph TD
    A["process pings the owner"] --> B["owner answers with ownerId/mode/build identity"]
    B --> C["compare the owner's build identity to this process's"]
    C -->|match| D["capable = true"]
    C -->|mismatch or missing| E["capable = false, reason recorded"]
    D --> F["requests forward normally"]
    E --> G["one early failure, with the reason"]
```

## Visual Proof

Not applicable. This repo's `create-pr.mjs` conservatively treats any diff touching `packages/app/src/main.ts` as UI-impacting, but the two lines changed there only add fields to an internal message-bus response and two `logger.warn`/`process.stderr.write` lines on the ownership-decision failure path — nothing rendered in a window, panel, or menu. There is no pixel to capture. (Because of this gate, the title/body on this PR were set with `gh pr edit` instead of `create-pr.mjs`.)

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `cd packages/app && pnpm exec vitest run src/__tests__/build-identity.test.ts`
- [x] `cd packages/app && pnpm exec vitest run src/__tests__/owner-build-mismatch.repro.test.ts`
- [x] `cd packages/app && pnpm test` (154 files, 1655 passed, 1 skipped — unchanged failure count)
- [x] `cd packages/app && pnpm run build` (bundling succeeds; confirmed the new fields are present in the compiled output)

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None — reverting restores the previous always-trusting owner-ping behavior with no data migration.
- Data migration? No

</details>

### DIFF
--- a/packages/app/src/__tests__/build-identity.test.ts
+++ b/packages/app/src/__tests__/build-identity.test.ts
@@ -1,0 +1,34 @@
+import { describe, it, expect } from 'vitest';
+
+import { describeBuildMismatch, type BuildIdentity } from '../build-identity.js';
+
+describe('describeBuildMismatch', () => {
+  it('is compatible when both sides report the same sha', () => {
+    const local: BuildIdentity = { version: '0.0.8', sha: 'abc1234' };
+    const remote: BuildIdentity = { version: '0.0.8', sha: 'abc1234' };
+    expect(describeBuildMismatch(local, remote)).toBeNull();
+  });
+
+  it('flags a mismatch when the shas differ, even if the version string matches', () => {
+    const local: BuildIdentity = { version: '0.0.8', sha: 'abc1234' };
+    const remote: BuildIdentity = { version: '0.0.8', sha: 'deadbee' };
+    const reason = describeBuildMismatch(local, remote);
+    expect(reason).not.toBeNull();
+    expect(reason).toContain('0.0.8 (deadbee)');
+    expect(reason).toContain('0.0.8 (abc1234)');
+  });
+
+  it('flags a remote that reports no build identity at all as incompatible', () => {
+    const local: BuildIdentity = { version: '0.0.8', sha: 'abc1234' };
+    const remote: BuildIdentity = { version: '', sha: '' };
+    const reason = describeBuildMismatch(local, remote);
+    expect(reason).not.toBeNull();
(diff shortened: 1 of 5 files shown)


EdbertChan:
@mergifyio queue

PR #7053, opened 2026-08-01
Time out and retry the post-task remote finalize step

EdbertChan:
## Summary

The post-task git record-and-push step opens a second SSH connection after the main task process has already exited. That connection previously had no time bound at all.

If a stalled remote credential prompt hung that connection, the task stayed "running" forever, with only a caller-side thirty-minute wait as a backstop.

This bounds each attempt and tries the step again up to three times before giving up, killing the stuck process each time it runs past its bound.

A task whose real work already finished no longer needs a full re-run just because one push attempt stalled.

## Review Claim

Approve bounding and repeating the post-task push step in the SSH executor so a stalled connection fails within minutes instead of staying stuck indefinitely.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

Every other execution-engine caller of the underlying primitive keeps its prior unbounded behavior; only the finalize step passes a bound, and the bound only fires after the given number of milliseconds elapses with no response.

## Slice Rationale

Kept to the one function this session traced a real stuck-task incident back to. Making every remote call visible when it fails is a separate, stacked change on top of this one.

## Non-goals

Does not change any other function that contacts the remote machine. Does not change the caller-side thirty-minute wait; that stays as an outer backstop.

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `cd packages/execution-engine && pnpm test -- ssh-executor` (41 passing, including 3 new cases proving the bound, the repeated attempts, and recovery on a later attempt)

</details>

## Revert Plan

<details>
<summary>Revert Plan</summary>

- Safe to revert? Yes.
- Revert command: `git revert <sha>`.
- Post-revert steps: None.
- Data migration? No.

</details>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of remote SSH command completion by detecting stalled operations and terminating them cleanly.
  - Added automatic retries for remote finalization, with up to three attempts before reporting failure.
  - Improved timeout error reporting and ensured failures are logged and surfaced consistently.
  - Preserved successful execution when a later retry completes the operation.

PR #7054, opened 2026-08-01
Log every remote-command failure, not just the finalize step

EdbertChan:
## Summary

Every function that contacts the remote machine funnels through one shared low-level primitive in the SSH executor.

Before this change, a failure there stayed silent by default. Nothing appeared anywhere unless an opt-in performance flag happened to be turned on.

This makes that one shared primitive report on its own, every time: a non-zero exit, a spawn failure, or the new time bound from the previous change firing.

Each report names the host, the calling step, why it ended, and a short tail of whatever the remote side printed.

That gives a future stuck-connection incident, like the one this session investigated, real evidence to start from.

## Review Claim

Approve unconditional failure reporting inside the one shared remote-contact primitive in the SSH executor, covering every caller instead of only the finalize step.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

The added reporting only runs on paths that already end the call unsuccessfully; the success path and every returned value are unchanged, and every existing caller keeps its own error-handling exactly as before.

## Slice Rationale

Kept separate from the timing change underneath it: that one bounds and repeats a single step, this one gives every step -- bounded or not -- a visible failure trail. Different concerns, same file region.

## Non-goals

Does not change what happens on success. Does not add a bound to any call that does not already have one.

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `cd packages/execution-engine && pnpm test -- ssh-executor` (43 passing, including 2 new cases proving a non-zero exit and a spawn failure both produce a report)

</details>

## Revert Plan

<details>

- Safe to revert? Yes.
- Revert command: `git revert <sha>`.
- Post-revert steps: None.
- Data migration? No.

</details>

### DIFF
--- a/packages/execution-engine/src/__tests__/ssh-executor.test.ts
+++ b/packages/execution-engine/src/__tests__/ssh-executor.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { EventEmitter } from 'node:events';
 import type { ChildProcess } from 'node:child_process';
 import { existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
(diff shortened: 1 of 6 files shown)


EdbertChan:
Mergify repair stopped: lowest open stack PR #7054 is based on `stack/EdbertChan/remote-finalize-timeout-retry/time-retry-post-task-remote-finalize-step--efb249b2`, which still belongs to open PR(s) #7053 outside this stack; leaving it alone to avoid dropping dependency changes

PR #7278, opened 2026-08-03
Bound LaunchDispatcher's ready-task top-up so a large burst can't starve the event loop

EdbertChan:
## Summary

Bulk-editing many pending Invoker tasks at once could make the app stop responding for a couple of seconds.

Editing 150 tasks made all of them ready to run at once. The scheduler's periodic check tried starting every one in a single pass, with nothing else running in between.

That pass took over two seconds. Anything else asked of the app during that window had to wait, and its reply window ran out.

The fix caps how many ready tasks that check starts per pass, reusing a limit the scheduler already applies elsewhere. Leftovers get picked up on the next pass, seconds later.

## Review Claim

`Orchestrator.startExecution()` accepts an optional cap on how many ready tasks it starts per call, and `LaunchDispatcher` now passes its existing per-tick limit into that cap.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

The cap is opt-in via a new optional parameter; every existing caller that omits it keeps today's exact unbounded behavior. The read/write sequence inside a single startExecution() call is unchanged — this only limits how many ready tasks enter that sequence per call, not the sequence itself. No new asynchronous steps or pauses are added anywhere.

## Slice Rationale

One scheduling change plus its accompanying tests and the two call-site fixes it depends on to actually take effect in the running app; a separate PR for just the new parameter would leave an intermediate state where nothing uses it yet.

## Non-goals

Does not change the existing per-tick lease-and-dispatch limit, which already had this kind of cap. Does not change how a single task's state gets written to disk. Does not add retries, timeouts, or backoff anywhere.

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `cd packages/app && pnpm test launch-dispatcher-topup-event-loop-lag.test.ts`
- [ ] `cd packages/app && pnpm test launch-dispatcher.test.ts`
- [ ] `cd packages/app && pnpm test runner-capacity-fill-guarantee.test.ts`
- [ ] `cd packages/app && pnpm test dispatch-capacity-invariants.test.ts`
- [ ] `cd packages/data-store && pnpm test`

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

</details>

### DIFF
--- a/packages/app/src/__tests__/dispatch-capacity-invariants.test.ts
+++ b/packages/app/src/__tests__/dispatch-capacity-invariants.test.ts
@@ -72,7 +72,7 @@ async function makeHarness(maxConcurrency = 4) {
     orchestrator: {
       prepareTaskForNewAttempt: (taskId, reason) => orchestrator.prepareTaskForNewAttempt(taskId, reason),
       getTask: (taskId) => orchestrator.getTask(taskId),
-      startExecution: () => orchestrator.startExecution(),
+      startExecution: (opts) => orchestrator.startExecution(opts),
       syncFromDb: (workflowId) => orchestrator.syncFromDb(workflowId),
       getTaskLaunchReadiness: (taskId) => orchestrator.getTaskLaunchReadiness(taskId),
     },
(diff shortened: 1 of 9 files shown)


EdbertChan:
@mergifyio queue

EdbertChan:
@mergifyio queue

PR #7279, opened 2026-08-03
[Disk Headroom Local Race](1) Guard local cleanup with task-state liveness, not process patterns

EdbertChan:
## Summary

The previous version of this change protected a directory from clearing by matching running processes' arguments against three known signatures.

That only covers exactly the tools it was written against -- any other provisioning step went unprotected, since it wasn't one of the three.

This replaces that guess entirely with a check against Invoker's own task state.

A directory is now protected only while some task's own record still points at it and that task is not yet finished.

That holds regardless of which tool, if any, happens to be running there at the moment, and it protects every kind of ongoing work, not just one setup step.

Protection is checked per directory entry, so one active task under a shared reclaimable directory only holds back its own subdirectory, not its siblings.

## Review Claim

Approve replacing a process-signature guess with a direct check against Invoker's own task state, so protection covers every kind of ongoing work by construction instead of only the specific tools someone happened to write a pattern for.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

Every task whose status is not `completed`, `closed`, or `stale` -- the same terminal set already used elsewhere in this codebase for this exact judgment -- has its workspace path protected from clearing, checked at both an exact-path match and any ancestor/descendant relationship, so a directory holding an active task's path is never cleared regardless of what else is or isn't running. A store read failure protects nothing rather than everything, since this worker's only job is emergency relief on already-critical disk pressure; failing toward "protect everything" could leave a host critically full indefinitely over one transient read error, a worse outcome than occasionally reclaiming one task's own workspace, which it can recreate. Covered by tests for: an active task's subdirectory surviving while its siblings still clear, a finished task's path not being protected, and a throwing store falling back to the old clear-everything behavior instead of crashing the sweep.

## Slice Rationale

One self-contained change: the new state-based check replaces the process-signature guess entirely in the same commit, since leaving both in place would mean maintaining two different, sometimes-disagreeing answers to the same question.

## Non-goals

Does not change how any other worker reads task state -- this reuses the exact same read shape (`listWorkflows`/`loadTasks`) already declared by another worker in this codebase, added to the shared dependency surface rather than invented fresh.

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `cd packages/execution-engine && pnpm test -- disk-headroom-reclaim`
- [ ] `pnpm run check:types`

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

</details>

### DIFF
--- a/packages/execution-engine/src/__tests__/disk-headroom-reclaim.test.ts
+++ b/packages/execution-engine/src/__tests__/disk-headroom-reclaim.test.ts
@@ -3,11 +3,15 @@ import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 
+import type { TaskState } from '@invoker/workflow-core';
+
 import {
   buildInvokerHomeCleanupScript,
   cleanupLocalInvokerHome,
+  computeProtectedLocalPaths,
   DISK_RECLAIMABLE_DIRS,
   DiskCleanupCooldownTracker,
+  type DiskHeadroomWorkerStore,
   isSafeInvokerHome,
   isSafeRemoteInvokerHomePath,
   resolveDiskCleanupCooldownMs,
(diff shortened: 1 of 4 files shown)


EdbertChan:

7279. #7279 — [Disk Headroom Local Race](1) Guard local cleanup with task-state liveness, not process patterns ← this PR
   Base: master
   Link: https://github.com/Neko-Catpital-Labs/Invoker/pull/7279

PR #7285, opened 2026-08-03
[Workflow-Resume FK Race](1) Route internal worker mutation submits through the missing-workflow helper

EdbertChan:
## Summary

The workflow-resume worker's wake tick reads pending-task state, then hands that off through the internal worker mutation channel to run again shortly after.

If the workflow it read was gone a moment before that hand-off runs, the underlying insert hits a foreign key that no longer exists.

The tick throws an uncaught error every single time this race happens -- and it happens often enough to fill production logs all day.

Two other paths already treat a workflow that vanished mid-mutation as a benign no-op, through one shared function built for exactly that judgment call.

The internal worker channel bypassed that shared function and called the coordinator directly instead, so it never got the same protection.

This change routes the internal worker channel through the same shared function, and adds its channel name to the function's allowlist of cases where a missing workflow is expected, not an error.

## Review Claim

Approve routing the internal worker mutation channel through the same shared "missing workflow is a benign no-op" function two other channels already use, instead of leaving it as the one channel that calls the coordinator directly and has no protection against this race.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

Every channel that already worked before this change behaves identically -- only the internal worker channel's call path changes, and only its behavior when a workflow has vanished changes (from an uncaught error to a benign no-op). Covered by a real repro in `workflow-mutation-submit.test.ts` using a real in-memory adapter and a real workflow deletion racing the mutation, plus a companion case proving the coordinator itself still throws the raw error when called directly, showing the race is genuine absent this fix.

## Slice Rationale

One self-contained change: the new call path and its allowlist entry land together with the test proving both are needed, since the fix isn't complete without either half.

## Non-goals

Does not add a second, local no-op judgment inside the workflow-resume tick itself -- every current and future internal worker shares the same mutation channel this fixes, so a local guard there would only cover one of them and would drift from the shared function over time.

## Visual Proof

`packages/app/src/main.ts` is flagged as UI-impacting by path alone, since it also owns window and menu creation. This diff only changes an internal mutation-submission call path; nothing under `packages/ui/` is touched. The GUI still renders normally with this diff applied:

GUI renders normally with this diff applied

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `cd packages/app && pnpm test -- workflow-mutation-submit`
- [ ] `pnpm run check:types`

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

</details>

### DIFF
--- a/packages/app/src/__tests__/workflow-mutation-submit.test.ts
+++ b/packages/app/src/__tests__/workflow-mutation-submit.test.ts
@@ -1,5 +1,7 @@
-import { describe, expect, it, vi } from 'vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { SQLiteAdapter } from '@invoker/data-store';
 import { submitWorkflowMutationOrAcknowledgeDeleted } from '../workflow-mutation-submit.js';
+import { PersistedWorkflowMutationCoordinator } from '../persisted-workflow-mutation-coordinator.js';
 
 function makeForeignKeyError(): Error {
   const error = new Error('FOREIGN KEY constraint failed') as Error & { errcode?: number; errstr?: string };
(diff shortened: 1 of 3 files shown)


EdbertChan:

7285. #7285 — [Workflow-Resume FK Race](1) Route internal worker mutation submits through the missing-workflow helper ← this PR
   Base: master
   Link: https://github.com/Neko-Catpital-Labs/Invoker/pull/7285

EdbertChan:
@mergifyio queue

PR #7413, opened 2026-08-04
this isnt rught [Workflow delete+resubmit UI drift](1) queue-status cache invalidation

EdbertChan:
## Summary

The orchestrator keeps a short-lived cache of queue status for the UI.

Two of its write paths never cleared this cache after running, so the UI could keep showing an old workflow set for a few seconds afterward.

This PR closes both gaps and adds a test that catches the drift.

## Review Claim

`createAndSync` and `deleteWorkflow` now null `queueStatusUiCache` on every call, matching the invalidation `writeAndSync` already performs on every other task-state write, so a `refresh:false` queue-status poll can never observe a workflow set that predates a just-completed `createAndSync` or `deleteWorkflow` call.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

Only cache invalidation is added; no read path, no cache TTL, and no other write path's behavior changes. Nulling the cache is always safe — the next `refresh:false` poll simply recomputes from `this.stateMachine.getAllTasks()`, which both `createAndSync` and `deleteWorkflow` already keep current via `stateMachine.restoreTask` and `syncAllFromDb`.

## Slice Rationale

One behavior slice — closing the two missing invalidation call sites for one already-diagnosed cache-timing defect — kept separate from a second, unrelated defect fixed in a later slice of this effort.

## Non-goals

No change to the cache TTL, to `writeAndSync`, to any read path, or to any other orchestrator method.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `cd packages/workflow-core && pnpm test -- queue-status-ui-cache-drift-repro`

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

</details>

### DIFF
--- a/packages/workflow-core/src/__tests__/queue-status-ui-cache-drift-repro.test.ts
+++ b/packages/workflow-core/src/__tests__/queue-status-ui-cache-drift-repro.test.ts
@@ -1,0 +1,133 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { Orchestrator } from '../orchestrator.js';
+import { InMemoryPersistence, InMemoryBus } from './helpers/cross-workflow-cascade-helpers.js';
+
+describe('queue status UI cache can drift from the DB', () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it('a newly-created queued task must not be invisible to the UI cache', () => {
+    const persistence = new InMemoryPersistence();
+    const orchestrator = new Orchestrator({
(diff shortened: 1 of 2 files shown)


EdbertChan:
Closing as duplicate of #7414. Both carry a byte-identical diff (same file hashes, single commit each, touching the same 2 lines in orchestrator.ts to invalidate queueStatusUiCache). Keeping #7414 since it has the fuller PR body explaining the user-facing report and stack context. Confirmed the underlying bug is real and still unfixed on master (repro test fails without the fix, 868 tests pass with it) — this isn't a redundancy close, just consolidating the duplicate.

PR #7418, opened 2026-08-04
[Fix headless delete bugs](1) Stop INVOKER_HEADLESS_STANDALONE leak

EdbertChan:
## Summary

`cleanElectronEnv()` strips several Electron-only variables from a copy of
`process.env` before a task's child process spawns. `INVOKER_HEADLESS_STANDALONE`
was missing from that set.

A parent process running in that mode leaked its own flag into every
spawned child, even children meant to run in the normal mode instead.

This change adds `INVOKER_HEADLESS_STANDALONE` to that same set of
variables `cleanElectronEnv()` already strips, so a spawned child's env no
longer carries the parent's standalone flag.

## Review Claim

Approve stripping `INVOKER_HEADLESS_STANDALONE` from the env object
returned by `cleanElectronEnv()`, with a regression test proving the flag
is absent from the stripped env while the parent's own `process.env` and
an unrelated var stay unchanged.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

The change is a single added `delete` statement on a key already handled by
the same deletion block as the other Electron-only variables, so it follows
an existing, already-reviewed pattern. `cleanElectronEnv()` still returns a
plain copy of `process.env`; the real `process.env.INVOKER_HEADLESS_STANDALONE`
in the parent process is left untouched, only the returned clone is affected.
The new test asserts both: the flag is absent from the clone, and the
parent's real env plus an unrelated var are unchanged.

## Slice Rationale

This is a single, atomic fix: one line added next to the other variables
`cleanElectronEnv()` already strips, together with the regression test
proving it, in the same file pair, with nothing else mixed in. Landing the
fix and its test in two slices would only add review overhead, not review
safety.

## Non-goals

Does not change any of the other variables `cleanElectronEnv()` already
strips (`ELECTRON_RUN_AS_NODE`, `ELECTRON_NO_ASAR`, `ELECTRON_NO_ATTACH_CONSOLE`,
`INVOKER_REPO_CONFIG_PATH`). Does not change `getEffectivePath()` or any other
env-resolution logic in `process-utils.ts`. Does not change how or when
`INVOKER_HEADLESS_STANDALONE` is set on the parent process itself.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `cd packages/execution-engine && pnpm test -- process-utils` (21/21 passed)

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

</details>

### DIFF
--- a/packages/execution-engine/src/__tests__/process-utils.test.ts
+++ b/packages/execution-engine/src/__tests__/process-utils.test.ts
@@ -121,6 +121,27 @@ describe('process-utils shell environment resolution', () => {
     expect(mockedSpawn).toHaveBeenCalledTimes(1);
   });
 
+  it('strips INVOKER_HEADLESS_STANDALONE from the cleaned child env', async () => {
+    const originalHeadlessStandalone = process.env.INVOKER_HEADLESS_STANDALONE;
+    const originalMarker = process.env.INVOKER_TEST_UNRELATED_MARKER;
+    try {
+      process.env.INVOKER_HEADLESS_STANDALONE = '1';
+      process.env.INVOKER_TEST_UNRELATED_MARKER = 'keep-me';
+
+      const { processUtils } = await loadProcessUtils();
+      const clean = processUtils.cleanElectronEnv();
+
+      expect(clean).not.toHaveProperty('INVOKER_HEADLESS_STANDALONE');
+      expect(clean.INVOKER_TEST_UNRELATED_MARKER).toBe('keep-me');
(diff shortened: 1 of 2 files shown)


EdbertChan:

7418. #7418 — [Fix headless delete bugs](1) Stop INVOKER_HEADLESS_STANDALONE leak ← this PR
   Base: master
   Link: https://github.com/Neko-Catpital-Labs/Invoker/pull/7418

PR #7560, opened 2026-08-05
[CI regression: a5d6b3e-playwright-launch-dispatch-stuck-lease](1) Remove duplicate launch-dispatch-stuck-lease shards from the Playwright CI matrix

EdbertChan:
## Summary

This keeps the launch-dispatch stuck-lease specs in one named Playwright shard.

The numbered shards keep their regular coverage without carrying that same focused shard workload.

## Review Claim

Approve the CI matrix inventory so the stuck-lease specs appear only in the dedicated Playwright shard.

## Review Lane

policy

## Review Unit

tooling-policy

## Safety Invariant

The change is limited to the Playwright workflow matrix and does not touch product code, spec code, or runner scripts.

## Slice Rationale

This is isolated to the Playwright matrix because the product code and spec files already have the intended coverage.

## Non-goals

- No product behavior change.
- No Playwright spec logic change.
- No runner script change.

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `node scripts/repro/repro-ci-playwright-shard-inventory.mjs`
- [ ] `node scripts/validate-pr-body-local.mjs --body-file <(gh pr view 7560 --repo Neko-Catpital-Labs/Invoker --json body --jq .body) --base master`

</details>

## Revert Plan

<details>
<summary>Revert Plan</summary>

- Safe to revert? Yes
- Revert command: `git revert 4a64a9c1d04851d52cd9eb501348bab0393e3239`
- Post-revert steps: None
- Data migration? No

</details>

PR #7652, opened 2026-08-06
Fix admin-bypass repair retry cap: record before submit, not after

EdbertChan:
## Summary

The admin-bypass repair pipeline hands a real Invoker workflow off, then writes a local ledger entry counting that attempt. It did those two things in the wrong order.

Tonight DO1's disk filled to 0 bytes free. The ledger write failed right after a repair workflow had already gone out. That workflow became invisible to the 3-attempt cap, so PR #7560's "PR Body" check kept getting relaunched every cycle instead of stopping — 11+ times and counting.

This swaps the order: record the attempt first, hand off the workflow second. A broken ledger write now blocks the handoff instead of silently orphaning it.

## Review Claim

Approve reordering `ledger.record()` before `submit_async_repair_plan()` at all three `AdminBypassRepairer` call sites, plus the new/updated tests proving the old order is unsafe.

## Review Lane

policy

## Review Unit

tooling-policy

## Safety Invariant

Pure Python, a two-line reorder repeated at three call sites plus test-only changes — no schema, API, or config changes. The full `mergify_admin_requeue` test suite (166 tests) passes. Worst case of the new ordering is a handoff that fails right after a successful record — that just overcounts one cap slot, which is safe; the old ordering's failure mode was an unbounded loop, which is not.

## Slice Rationale

The ordering fix and its test coverage are inseparable — landing one without the other either ships an unverified behavior change or adds tests for a bug that's already fixed. Kept as a single slice.

## Non-goals

Does not address why DO1's disk filled up (root-caused and fixed separately: 44 stale hourly DB backups, ~25GB reclaimed, the existing Disk Hygiene Reaper worker now runs unblocked). Does not add alerting/observability for ledger write failures beyond making them fail closed.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `python3 -m unittest scripts.test_mergify_admin_requeue scripts.test_mergify_admin_requeue_repairer` — 166 tests, all pass
- [x] `python3 -m unittest scripts.test_mergify_admin_requeue_repairer` in isolation, run against pre-fix code first — one of two tests fails (`test_a_broken_ledger_write_blocks_submission_instead_of_orphaning_it`), proving the bug; passes after the fix

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Revert command: `git revert 3c346068d`
- Post-revert steps: None
- Data migration? No

</details>

### DIFF
--- a/scripts/mergify_admin_requeue_repairer.py
+++ b/scripts/mergify_admin_requeue_repairer.py
@@ -193,8 +193,13 @@ class AdminBypassRepairer:
             head_sha=start_head,
             plan_name=plan.plan_name,
         )
-        async_repair.submit_async_repair_plan(plan)
+        # Record before submitting: once submitted, the workflow is real and
+        # running whether or not this process survives another instruction.
+        # Recording first means a broken ledger write (e.g. disk full) raises
+        # here and skips the submission entirely, instead of leaving a real,
+        # running repair permanently invisible to the retry-cap count.
         self.ledger.record("repair-check", pr.number, start_head, check_name, now)
+        async_repair.submit_async_repair_plan(plan)
         return self.blocked_outcome("submitted", check_name, start_head, start_head)
 
     def repair_conflict(self, pr: PrSnapshot, reason: str, now: int | None = None) -> RepairOutcome:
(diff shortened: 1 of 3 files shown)


EdbertChan:
Mergify repair stopped: missing required check PR Body

EdbertChan:
@mergifyio queue

PR #7712, opened 2026-08-07
[Bugfix: harden stale launch-dispatch claim fencing guard](1) Name the stale-candidate fencing guard

EdbertChan:
## Summary

A launch-dispatch claim must only lease a row whose task still matches the expected attempt and generation. A stale row must be abandoned instead.

This already happens today. The claim step checks four conditions before leasing: the task exists, its status still allows a claim, and its attempt id and generation still match.

This change does not touch that behavior. It only names the four checks as one function, `isLaunchDispatchCandidateStale`, so each condition can be checked on its own.

Every row that passed or failed these checks before still passes or fails them the same way now.

## Review Claim

`packages/data-store/src/sqlite-adapter.ts`'s `claimLaunchDispatchAtomic` already abandons a stale row before leasing it when the joined task is missing, not claimable, or has a mismatched attempt id or execution generation. This PR names that same check as one function, `isLaunchDispatchCandidateStale`, called from the same place in the claim step, with identical results for every input.

## Review Lane

behavior

## Review Unit

validation-policy

## Safety Invariant

For every candidate row, the claim transaction still abandons and skips a row exactly when today's four staleReason conditions hold (missing task, non-claimable status, attempt mismatch, generation mismatch), and still leases a row exactly when none of them hold. No condition is added, removed, or reordered — the new function reproduces the same four `if`/`else if` branches in the same order and returns the same string values.

## Slice Rationale

This is a single hardening step: name the four existing stale checks as one function and call it from the one place that already runs them. Nothing else changes in this diff.

## Non-goals

- No change to the abandoned-row `UPDATE` statement or its bind parameters.
- No change to any staleReason string value.
- No change to candidate scan order or to which candidate gets leased.
- No new caller of the new function beyond the existing claim step and its direct unit tests.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `cd packages/data-store && pnpm test -- -t 'abandons a stale selected-attempt candidate and scans to the next valid row'` — ran the full `sqlite-adapter.test.ts` suite (290 tests) plus the rest of the package (442 tests total across 24 files); all 442 passed, 0 failed.
- [x] New `isLaunchDispatchCandidateStale` unit tests (missing task, non-claimable status, mismatched attempt, mismatched generation, fully-matching candidate) pass as part of the same run.

</details>

## Revert Plan

<details>

- Safe to revert? Yes.
- Revert command: `git revert <sha>`.
- Post-revert steps: None — the inline checks return with identical behavior.
- Data migration? No.

</details>

### DIFF
--- a/packages/data-store/src/__tests__/sqlite-adapter.test.ts
+++ b/packages/data-store/src/__tests__/sqlite-adapter.test.ts
@@ -2,7 +2,7 @@ import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import { mkdtempSync, rmSync, statSync, existsSync, readdirSync } from 'node:fs';
 import { join } from 'node:path';
 import { tmpdir } from 'node:os';
-import { SQLiteAdapter } from '../sqlite-adapter.js';
+import { SQLiteAdapter, isLaunchDispatchCandidateStale } from '../sqlite-adapter.js';
 import type { Workflow, Conversation, WorkerActionWrite, TerminalSessionRecord, InAppPlanningSessionRecord } from '../adapter.js';
 import { createAttempt } from '@invoker/workflow-core';
 import type { Attempt, TaskState, TaskStateChanges } from '@invoker/workflow-core';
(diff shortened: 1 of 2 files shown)


EdbertChan:

7712. #7712 — [Bugfix: harden stale launch-dispatch claim fencing guard](1) Name the stale-candidate fencing guard ← this PR
   Base: stack/EdbertChan/plan/bugfix-harden-cascading-fk-delete-ordering-guard/bugfix-harden-cascading-fk-delete-ordering-guard--60b29583
   Link: https://github.com/Neko-Catpital-Labs/Invoker/pull/7712

EdbertChan:
@mergifyio queue

PR #7738, opened 2026-08-07
[Stop the async repair normalize task from blocking on its own bytecode cache](1) Make the normalize dirty-gate immune to its own bytecode cache

EdbertChan:
## Summary

Every admin-bypass-repair-check workflow was failing at its normalize step with `blocked_dirty: repair left the working tree dirty`, even when the repair agent had correctly committed nothing.

The script's own imports made CPython write `scripts/__pycache__/` into the fresh task checkout. That directory was not gitignored, so the dirty check tripped on the script's own bytecode.

This PR stops the normalize process from writing bytecode at all. The script sets `sys.dont_write_bytecode` before importing its siblings, and the generated plan command becomes `python3 -B`.

The `-B` flag matters on its own: the command text renders from the cron host's master checkout, while the executed script comes from the PR branch under repair.

That means `-B` immediately fixes repair runs against branches cut before this change.

`.gitignore` also gains `__pycache__/`, so any Python the repair agent itself runs cannot dirty a task checkout either. A genuine `blocked_dirty` failure now names the offending `git status --porcelain` entries.

## Review Claim

The normalize dirty-gate can no longer be tripped by CPython bytecode written by the normalize process itself (the script sets `sys.dont_write_bytecode`, the generated invocation adds `-B`, and `__pycache__/` is gitignored), and a genuine blocked_dirty failure now names the offending `git status --porcelain` entries instead of an unactionable one-liner.

## Review Lane

policy

## Review Unit

tooling-policy

## Safety Invariant

Three narrow edits with no control-flow change on the success path: one interpreter flag set before imports in `scripts/mergify_admin_requeue_repair_normalize.py` plus an enriched stderr string in its existing dirty branch, one literal `-B` added to the generated command string in `scripts/mergify_admin_requeue_async_repair.py`, and one ignore rule appended to `.gitignore`. No schema, API, planner-gate, or repairer change; the two touched test files gain cases only. Nothing currently tracked in the repo matches `__pycache__/` (`git ls-files | grep __pycache__` is empty), so the new ignore rule cannot untrack or hide existing content.

## Slice Rationale

[...]

PR #7651, opened 2026-08-06
[Promote Fast Required Checks](1) Promote 4 fast core state-machine checks to required

EdbertChan:
## Summary

Four core checks already ran on every PR and master push, but a red result never blocked anything from landing.

They cover task/workflow dispatch and state-machine behavior, take about a minute each, and have been green for 5 consecutive master runs.

This adds them to Mergify's required list. A fifth check in the same class runs 6-13 minutes per run, so it stays out for now.

## Review Claim

Approve adding four green, ~1-minute checks to Mergify's `merge_conditions`, so a regression in core dispatch/state-machine behavior blocks the queue instead of merging silently.

## Review Lane

policy

## Review Unit

tooling-policy

## Safety Invariant

Pure config: four `check-success` lines added to an existing list. No script changed. All four checks already run today with no schedule or trigger change — this only changes whether their result blocks a merge. Confirmed locally: all four pass (see Test Plan), and `.mergify.yml` validates against Mergify's schema.

## Slice Rationale

One conceptual change — narrowing the required-check gap for core behavior — landed as a single config diff.

## Non-goals

- Does not add the slower fifth check in this class (6-13 min); that's a separate cost/benefit call.
- Does not change anything about the checks that stay non-required (Vitest Workspace, e2e-proof, Playwright) — those remain red-on-master and out of scope here.
- Does not modify any check's own logic or script content.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `mergify config validate` — configuration file is valid
- [x] `python3 -c "import yaml; yaml.safe_load(open('.mergify.yml'))"` — parses, required-check list confirmed
- [x] `bash scripts/test-suites/required/17-merge-gate-concurrency-repro.sh` — exit 0
- [x] `bash scripts/test-suites/required/25-launch-dispatch-queue-handoff-repro.sh` — exit 0
- [x] `bash scripts/test-suites/required/18-start-running-mece-repros.sh` — exit 0
- [x] `bash scripts/test-suites/required/16-branch-carry-forward.sh` — exit 0
- [x] All four checks green across the last 5 completed master CI runs (checked via `gh run view --json jobs`)

</details>

## Revert Plan

<details>
<summary>Revert Plan</summary>

- Safe to revert? Yes
- Revert command: `git revert <sha>`
- Post-revert steps: None. Reverting restores the four checks to non-blocking.
- Data migration? No

</details>

PR #7782, opened 2026-08-07
Run required-fast-extra on the Mergify merge-queue ref, not around it

EdbertChan:
PR #7651 promoted required-fast-extra's five checks (Branch Carry
Forward, Merge Gate Concurrency Repro, Start Running MECE Repros,
Launch Dispatch Queue Repro, PR Babysit Harness) to required in
.mergify.yml's shared merge_conditions. But the job's own `if`
excludes mergify/merge-queue/* refs, and its `needs: build-artifacts`
dependency only runs on push/schedule/manual/merge-queue events -
never on an ordinary PR ref. The two conditions are mutually
exclusive with every context a not-yet-merged PR can be in: build-
artifacts skips on the PR's own branch, and required-fast-extra skips
on Mergify's speculative merge-queue branch. The check-success
conditions can only ever be satisfied by a commit already on master,
after the PR that needed them has already merged - a deadlock that
blocks every admin-bypass and default-queue PR permanently in
'checks running', confirmed via #7738 stuck 45+ minutes with every
other required check green.

Flip the guard to match required-fast and build-artifacts: run on
Mergify's merge-queue ref (in parallel with required-fast, per the
original comment's intent), not around it.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -380,7 +380,7 @@ jobs:
         run: ${{ matrix.command }}
 
   required-fast-extra:
-    if: ${{ github.event_name != 'pull_request' || !startsWith(github.head_ref, 'mergify/merge-queue/') }}
+    if: ${{ github.event_name != 'pull_request' || startsWith(github.head_ref, 'mergify/merge-queue/') }}
     needs: build-artifacts
     runs-on:
       labels: ${{ matrix.runner_label }}


EdbertChan:
Duplicate of #7780, which already landed the identical fix (flip required-fast-extra's if-condition to run on the Mergify merge-queue ref). Closing in favor of the merged PR.

PR #7957, opened 2026-08-08
Bugfix: infra-repair worker recreates a task forever on a persistent missing-worktree failure

EdbertChan:
## Summary

The infra-repair worker recreates a failed SSH task when its remote worktree is missing.

If the underlying problem is persistent, it recreated the same task forever, once per tick, with no limit.

This happened in production on `remote_digital_ocean_1`: one task was recreated dozens of times over 14 hours.

The cause is a dedup key that includes the task's generation and version. Both change every recreate, so the guard never caught a repeat.

This repo already has a shared, generation-independent attempt counter built for this exact failure shape. The infra-repair worker was never wired into it.

This change wires it in, so infra-repair now stops after a configured number of attempts (default 3) and leaves a note for an operator instead of looping.

## Review Claim

`submitFollowUpMutation` in `infra-repair-worker.ts` now checks a shared per-task attempt counter before queuing a follow-up task mutation, and advances that counter afterward, so repeated recovery attempts across task recreates are durably bounded instead of unbounded.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

The attempt counter only gates the four existing recovery paths' follow-up queuing point; it does not change failure classification, target selection, or the OAuth-session-expired alert path (which already never acts automatically). When the counter is exhausted, the worker records a decision explaining why instead of silently dropping the failure, so it stays visible via `query worker-decisions`.

## Slice Rationale

This is the minimal fix for one confirmed production incident (an unbounded recreate loop on `remote_digital_ocean_1`). It reuses an existing, already-tested counter (`checkAutoFixRetryCap` / `recordAutoFixRetryConsumed`) rather than inventing a new mechanism.

## Non-goals

Does not change the separate cooldown mechanism used by the env-repair/mirror-repair remote-script paths (`runTargetRepairWithCooldown`), which already works correctly. Does not add a new config knob -- it reuses `deps.autoFix.defaultAutoFixRetries`, the same source the autofix worker already reads from.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `cd packages/execution-engine && pnpm test -- infra-repair-worker.test.ts`
- [x] `cd packages/execution-engine && pnpm test`

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Revert command: `git revert b34cf68`
- Post-revert steps: None
- Data migration? No

</details>

### DIFF
--- a/packages/execution-engine/src/__tests__/infra-repair-worker.test.ts
+++ b/packages/execution-engine/src/__tests__/infra-repair-worker.test.ts
@@ -73,7 +73,10 @@ function toRecord(write: WorkerActionWrite): WorkerActionRecord {
   };
 }
 
-function makeHarness(tasksInput: TaskState[] = [makeTask()]) {
+function makeHarness(
+  tasksInput: TaskState[] = [makeTask()],
+  options: { defaultAutoFixRetries?: number } = {},
+) {
   const tasks = new Map(tasksInput.map((task) => [task.id, task]));
   const actions = new Map<string, WorkerActionRecord>();
   const submissions: Array<{ workflowId: string; priority: WorkflowMutationPriority; channel: string; args: unknown[] }> = [];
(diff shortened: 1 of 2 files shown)


EdbertChan:
@mergifyio queue

PR #7949, opened 2026-08-08
[Hardening: WorkResponse.outputs array rejection gets a named guard](1)[REFACTOR: Extract Method] isValidWorkResponseOutputs guard

EdbertChan:
## Summary

Today `validateWorkResponse` inline-checks that `outputs` is a non-array object, guarding against a silent-corruption bug from commit ba2ab851a.

That inline check now lives in its own small, named, exported function: `isValidWorkResponseOutputs`.

`validateWorkResponse` calls the new function from the exact same place, with the exact same error text and result. No caller-visible behavior changes.

This is a hardening/testability slice, not a bugfix. The array-rejection invariant is already correct and already covered by a regression test.

This slice only makes the specific condition directly testable on its own.

## Review Claim

The inline `!r.outputs || typeof r.outputs !== 'object' || Array.isArray(r.outputs)` condition inside `validateWorkResponse` becomes a standalone, named, exported boolean function called from the same place with the same result.

## Review Lane

refactor

## Review Unit

contract

## Safety Invariant

`validateWorkResponse` still returns the same error for every input; behavior is unchanged before and after this slice, for every existing caller and every existing test. No behavior change.

## Slice Rationale

One slice: give the existing inline condition its own name and export it, plus a direct unit test, nothing else. The three sibling copies of this check in `protocol`, `core`, and `workflow-core` are out of scope for this slice.

## Non-goals

This slice makes no behavior change: no new call sites, no change to `validateWorkResponse`'s own signature or error text, no unrelated cleanup, no documentation edits, and no change to the three sibling duplicate checks in `protocol`, `core`, or `workflow-core`.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `cd packages/contracts && pnpm test -- -t 'rejects array outputs (silent-corruption guard)'`

</details>

## Revert Plan

<details>
<summary>Revert Plan</summary>

- Safe to revert? Yes
- Revert command: `git revert <sha>`
- Post-revert steps: None
- Data migration? No

</details>

PR #7962, opened 2026-08-08
Bugfix: Extract dependency-branch-metadata guard into named function

EdbertChan:
invoker: wf-1785491684476-24/fix-branch-persisted-at-completion — Extract the duplicated completed-dependency branch-metadata check into a named, exported guard function and add a direct unit test for it.

Review claim: This is not a bugfix. The invariant (a completed dependency must carry branch metadata) is already enforced by two near-identical inline throw blocks in packages/execution-engine/src/task-runner-prepare.ts:55-71 and is already proven end-to-end by task-runner.test.ts:1864. This slice only extracts the duplicated inline logic into `assertCompletedDependencyHasBranch()`, calls it from both existing call sites with byte-identical error messages, and adds a direct unit test for the extracted function.

Review lane: behavior

Safety invariant: For every (taskId, depLabel, dep) input, the extracted function throws in exactly the same cases and with exactly the same message text as the two inline blocks it replaces; no other behavior in buildWorkRequest changes.

Slice rationale: One extraction-plus-test slice: pull the two duplicated inline checks out into a single named function and directly test it, nothing else changes.

Non-goals: No change to the onBranchResolved early-write path in this same file, no change to transitions.ts, no new fields, no doc edits.

Acceptance criteria:
- `cd packages/execution-engine && pnpm test -- -t "fails task when a completed worktree dep has no branch"` exits 0 after the change.

### DIFF
--- a/packages/contracts/src/__tests__/validation.test.ts
+++ b/packages/contracts/src/__tests__/validation.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest';
-import { validateWorkRequest, validateWorkResponse } from '../validation.js';
+import { validateWorkRequest, validateWorkResponse, isValidWorkResponseOutputs } from '../validation.js';
 import { createWorkRequest } from '../types.js';
 import type { WorkResponse } from '../types.js';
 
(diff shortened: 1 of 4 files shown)


EdbertChan:
Closing as redundant. Rebasing this branch onto current `origin/master` shows master already ships the identical extraction via #7981 (`da650d03f`, "[Bugfix: Extract dependency-branch-metadata guard into named function]... Combine duplicated branch-metadata checks into one function"):

- `packages/execution-engine/src/task-runner-prepare.ts` on master already exports `assertCompletedDependencyHasBranch(taskId, depLabel, dep: TaskState | undefined)` — functionally equivalent to (and slightly more defensive than) this PR's version, which requires a non-undefined `dep` and pushes the undefined-check to call sites.
- `packages/execution-engine/src/__tests__/task-runner.test.ts` on master already has direct unit tests for `assertCompletedDependencyHasBranch`, including a case this PR doesn't cover (`dep` is `undefined`).

This PR's branch also still carried an already-landed, unrelated commit (`isValidWorkResponseOutputs` in `packages/contracts/src/validation.ts`, shipped via #7949) in its history, which was showing up as unrelated diff-bleed in the live diff — a separate symptom of the same staleness, not a reason to keep this PR open on its own.

No remaining net-new behavior after reconciling with master. Closing without merging.

PR #8116, opened 2026-08-09
Fail fast with a clear error when the IPC socket path is too long

EdbertChan:
## Summary

`./run.sh --headless run <plan>` hung for 60 seconds, then failed with "could not reach a standalone shared owner" — even while a real owner process was running the whole time.

The real cause: `INVOKER_DB_DIR` was long enough that the derived Unix socket path exceeded the OS's ~104-character limit. Binding or connecting to it failed with `EINVAL`.

The code treated that `EINVAL` the same as "server may just be booting," and retried silently forever, both connecting and binding. It's permanent, not transient.

This adds a narrow classifier for that one error class and records it once. The headless CLI's bootstrap loop checks it and fails fast with a clear message, instead of exhausting its 60-second timeout.

## Review Claim

`IpcBus` now records a `getFatalConnectError()` when a connect or bind attempt hits `EINVAL`/`ENAMETOOLONG` (socket path too long), and the headless CLI's bootstrap loop checks it to fail fast with the real cause instead of waiting out its timeout.

## Review Lane

behavior

## Review Unit

activation-surface

## Safety Invariant

Every existing retry path (`ECONNREFUSED`, `ENOENT`, `EADDRINUSE`, and all other errors) keeps its exact prior behavior — `resolveReady()`/`scheduleRecovery()` still fire unconditionally on every path. The only new behavior is recording the fatal error alongside the existing retry, and one new early-exit check in the headless bootstrap loop. No existing caller of `bus.ready()` can observe a rejection that wasn't possible before; that promise's resolve/reject contract is unchanged.

## Slice Rationale

Single cohesive change: the transport-level classification and the one call site that needed it (headless CLI bootstrap, the only place this was actually observed hanging). Both `## Review Unit` checks pass together with no product/policy/proof file split needed.

## Non-goals

- Does not change how `main.ts`'s owner-side startup paths use `bus.ready()` — deliberately left alone given 11+ existing call sites depend on it always resolving.
- Does not auto-shorten or relocate the socket path. The fix is "fail fast with a clear message," not "silently work around it" — the user still needs to shorten `INVOKER_DB_DIR` or set `INVOKER_IPC_SOCKET`, they just find out in under a second instead of a minute.
- Does not touch the `EADDRINUSE` server-election race logic.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `cd packages/transport && pnpm test` — 53 tests passed (51 pre-existing + 2 new), including two new tests proving `getFatalConnectError()` fires for both a server (`allowServe: true`) and a pure client (`allowServe: false`) on a too-long socket path.
- [x] `cd packages/app && pnpm exec vitest run src/__tests__/headless-client.test.ts src/__tests__/headless-client-fixture-cleanup.test.ts` — 40 passed, 2 pre-existing skips, no regressions.
- [x] Real end-to-end repro, before this fix (pre-fix commit, same 154-char `INVOKER_DB_DIR` that originally triggered this): hangs the full 60s, then fails with the generic error, while `ps aux` confirms a real owner process was alive the whole time.
- [x] Real end-to-end repro, after this fix, same exact command and path:
  ```
  [headless-client] bootstrap fatal connect error attempts=1 elapsedMs=3: Cannot use IPC socket path
  "/private/tmp/.../invoker-15wf-demo/db/ipc-transport.sock" (154 chars): EINVAL. Unix domain socket
  paths are limited to roughly 100 characters on most operating systems. Set INVOKER_IPC_SOCKET to a
  shorter path, or shorten INVOKER_DB_DIR.
  Error: Cannot use IPC socket path ... (154 chars): EINVAL. ...
  ```
  Total wall time: 0.52s (was 60s+).
- [x] Real end-to-end repro of the happy path (short socket path) after this fix: unaffected — workflow submitted, task ran, completed, exit 0, in ~1.2s to owner-ready as before.

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

</details>

### DIFF
--- a/packages/app/src/headless-client.ts
+++ b/packages/app/src/headless-client.ts
@@ -555,6 +555,13 @@ async function ensureStandaloneOwnerViaBootstrap(bus: MessageBus): Promise<void>
         );
         return;
       }
+      if (bus instanceof IpcBus) {
+        const fatal = bus.getFatalConnectError();
+        if (fatal) {
+          delegationClientLog(`bootstrap fatal connect error attempts=${attempts} elapsedMs=${Date.now() - startedAt}: ${fatal.message}`);
+          throw fatal;
+        }
+      }
       await new Promise((resolveDelay) => setTimeout(resolveDelay, 200));
     }
     delegationClientLog(`bootstrap timeout elapsedMs=${Date.now() - startedAt}`);
--- a/packages/transport/src/__tests__/ipc-bus.test.ts
+++ b/packages/transport/src/__tests__/ipc-bus.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect, afterEach } from 'vitest';
 import { join, dirname } from 'node:path';
-import { chmodSync, existsSync, mkdtempSync } from 'node:fs';
+import { chmodSync, existsSync, mkdirSync, mkdtempSync } from 'node:fs';
 import { homedir, tmpdir } from 'node:os';
 import { spawn } from 'node:child_process';
 import { fileURLToPath } from 'node:url';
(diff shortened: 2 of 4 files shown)


EdbertChan:

PR #7222, opened 2026-08-03
[Bugfix: clicking empty background on the workflow graph dismisses mini-DAG](1) Restore no-op background click handler (regression reintroduced after PR #4982)

EdbertChan:
## Summary

Clicking on empty background in the workflow graph used to clear your selected task and workflow. It also dismissed the mini-DAG panel for the selected workflow.

PR #4982 fixed this: background clicks became a no-op except for closing an open context menu.

PR #5067 silently reverted that fix two days later, and flipped the one test that would have caught it.

This PR restores the #4982 no-op behavior and un-flips that test back to asserting the mini-DAG panel survives a background click.

CI also caught a second existing test, `invoker-terminal.test.tsx`. It used a background click to set up a "cleared selection" scenario.

That setup step no longer works once background click is a no-op. This PR updates it: select a task node directly, then dismiss via Escape.

## Review Claim

`handleDagSurfaceClick` in `packages/ui/src/App.tsx` must close an open context menu on background click and otherwise be a no-op. It must not clear `selectedTaskId`, `selectedWorkflowId`, or set `workflowSelectionDismissed`. Separately, `invoker-terminal.test.tsx`'s "preserves a cleared graph selection" test must clear selection without relying on background click, and still pass.

## Review Lane

behavior

## Review Unit

activation-surface

## Safety Invariant

All other behavior in `packages/ui/src/App.tsx` stays identical; only `handleDagSurfaceClick`'s body changes. Node/edge click handling and context-menu-close-on-background-click are unaffected. The three touched tests only update assertions and setup steps to match the restored no-op behavior; none of them add new product behavior.

## Slice Rationale

One behavior slice: the click handler defect and the three existing tests that cover it, nothing else. An automated end-to-end visual-proof test for this exact interaction ships as the next PR in this stack, since e2e/visual proof is reviewed as its own unit separate from the product change it proves.

## Non-goals

No refactor, no new fields, no unrelated cleanup, no doc edits. Does not change context-menu-close-on-background-click behavior, node/edge click handling, or any other click handler.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `cd packages/ui && pnpm test -- src/__tests__/home-workflow-click-mini-dag-dismiss-repro.test.tsx src/__tests__/task-interaction.test.tsx src/__tests__/invoker-terminal.test.tsx`
- [x] `cd packages/ui && pnpm test` — `830 passed` (full Vitest suite, 88 files, 0 failed)

</details>

## Visual Proof

[...]

PR #8130, opened 2026-08-09
[DAG background-click camera zoom](3) Fix finishPanePan clearing the viewport transform on a plain click

EdbertChan:
## Summary

While proving the background-click regression fix at 15-workflow scale, the actual video (not just a DOM assertion) showed a real, separate camera-zoom bug on every click of empty background, on `origin/master`, unrelated to the already-fixed selection-clearing regression.

`finishPanePan()` clears the graph's viewport CSS transform on every pane click, assuming React Flow's own re-render will immediately replace it. For a plain click (no drag), the target viewport equals the current one, so `setViewport()` is a no-op — React Flow never re-renders, and the transform stays cleared. Nodes then render at raw, untransformed positions, which looks exactly like an unwanted zoom.

Traced with a MutationObserver and a patched style setter to get a real stack-level answer instead of guessing: the clear happens synchronously inside the pane's mouseup handler, and never recovers.

## Review Claim

`finishPanePan()` in `packages/ui/src/components/WorkflowGraph.tsx` only clears the viewport's inline transform when the pan actually changed position (`pan.hasMoved`); for a plain click, it re-applies the unchanged transform instead of clearing it.

## Review Lane

behavior

## Review Unit

activation-surface

## Safety Invariant

The real-drag path is untouched: `pan.hasMoved` is only true after `updatePanePanViewport` ran at least once, and that path still calls `clearViewportInlineTransform` exactly as before, so React Flow's reactive re-render still takes over cleanly after a genuine drag. Only the zero-movement (click) path changes, from clearing to re-applying the same, already-correct transform.

## Slice Rationale

Isolated to one function in one file. Stacked on top of the guard slice since it's an unrelated fix found while checking that slice's proof; the test that actually asserts this fix is its own slice next, since a proof-lane test file and an activation-surface product file can't ship in the same PR per this repo's review-unit rules.

## Non-goals

- Does not touch `handleDagSurfaceClick` or any selection-clearing logic (already fixed by #7222, unrelated).
- Does not touch the real-drag pan path.
- Does not add the strengthened test assertion — that's the next slice, once this fix is in place to make it pass for real instead of via an expected-fail marker.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `cd packages/ui && pnpm test` — 838 tests passed across 89 files, no regressions (includes `workflow-graph.test.tsx`, `graph-camera.test.ts`, `browser-surface-camera-resnap.test.tsx`, and two graph-jump/disappear repro suites).
- [x] Real end-to-end proof: `npx playwright test e2e/visual-proof.spec.ts -g "dag-full-graph-bg-click-noop"` — before this fix, `after-bg-click="none"` (transform wiped); after this fix, `after-bg-click="translate(318.5px, 265.5px) scale(0.3)"`, identical to `after-select`. Ran 3 times, all consistent.
- [x] Watched the actual video frame-by-frame (not just the assertion) both before and after this fix to confirm visually, not just via DOM string comparison.

</details>

## Visual Proof

Real Electron build, 15 real seeded workflows, real mouse click on empty background — captured via `scripts/ui-visual-proof.sh` with `CAPTURE_VIDEO=1`.

**Before this fix (still on master today)** — 15 compact workflow cards collapse to 3 large ones the instant background is clicked, even with selection preserved (the earlier, separate bug is already fixed — this is a second, independent issue):

[dag-camera-zoom-BEFORE-walkthrough.webm](https://pub-cf646cda2bd945d683792ee19b5f9d98.r2.dev/pr-images/20260809-7b5c40/dag-camera-zoom-BEFORE-walkthrough.webm)

dag-camera-zoom-before-click
dag-camera-zoom-after-click-BUGGY

**After this fix** — same click, camera stays exactly where it was:

[dag-camera-zoom-AFTER-walkthrough.webm](https://pub-cf646cda2bd945d683792ee19b5f9d98.r2.dev/pr-images/20260809-7b5c40/dag-camera-zoom-AFTER-walkthrough.webm)

dag-camera-zoom-after-click-FIXED

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workflow graph interactions by correctly distinguishing clicks from drags.
  * Preserved the graph’s position after clicks and ensured completed pans display at the correct viewport.

### DIFF
--- a/packages/ui/src/__tests__/workflow-graph.test.tsx
+++ b/packages/ui/src/__tests__/workflow-graph.test.tsx
@@ -735,6 +735,31 @@ describe('WorkflowGraph', () => {
     expect((viewport as HTMLElement).style.transform).toBe('');
   });
 
+  it('keeps pane-pan inline transforms when a drag ends at the starting viewport', async () => {
+    getViewportMock.mockReturnValue({ x: 10, y: 20, zoom: 0.75 });
+
+    await renderAndSettleInitialFit({
+      workflows: new Map([['wf-a', wf('wf-a', 'running')]]),
+      selectedWorkflowId: 'wf-a',
+      statusFilters: new Set(),
+      onSelectWorkflow: () => {},
+      onWorkflowContextMenu: () => {},
+    });
+
+    const viewport = screen.getByTestId('workflow-graph-react-flow').querySelector('.react-flow__viewport');
+    expect(viewport).not.toBeNull();
+
+    const pane = screen.getByTestId('rf__pane');
+    fireEvent.mouseDown(pane, { clientX: 100, clientY: 100, button: 0 });
+    fireEvent.mouseMove(window, { clientX: 130, clientY: 88, buttons: 1 });
+    await waitFor(() => expect((viewport as HTMLElement).style.transform).not.toBe(''));
+    fireEvent.mouseMove(window, { clientX: 100, clientY: 100, buttons: 1 });
+    fireEvent.mouseUp(window, { clientX: 100, clientY: 100, button: 0 });
+
+    expect(setViewportMock).toHaveBeenCalledWith({ x: 10, y: 20, zoom: 0.75 }, { duration: 0 });
+    expect((viewport as HTMLElement).style.transform).toBe('translate(10px, 20px) scale(0.75)');
+  });
+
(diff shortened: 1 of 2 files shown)


EdbertChan:

PR #8136, opened 2026-08-09
Add a mechanical gate requiring proof of manual inspection in PR bodies

EdbertChan:
## Summary

A captured screenshot or video was never proof that anyone opened it and looked.

CI now checks for that directly: a `## Visual Proof` section with media but no `Manually inspected:` line describing what was actually seen now fails the check.

This adds one shared page, `skills/prove-it`, stating the rule once. `make-pr`, `visual-proof`, and `land-stack` each get a short pointer to it plus the one instruction specific to that skill — a fresh status query for `land-stack`, actually opening media for the other two.

## Review Claim

The PR-body checker now requires a `Manually inspected:` line in `## Visual Proof` whenever visual proof media is present; `skills/prove-it` is the one shared statement of the rule, referenced from three other skills.

## Review Lane

policy

## Review Unit

tooling-policy

## Safety Invariant

The new check only fires when `requiresVisualProof` is already true and only adds a third `else if` branch after the two existing checks, so a body that already fails for missing media or missing animation still reports that error first — nothing about the existing two checks changes. Every skill instruction added here has a matching `must_contain` assertion in its own contract test script, so removing the instruction text fails CI.

## Slice Rationale

Kept apart from the two prior slices: the prose-only rules classify under a different Review Unit than this one, and CI does not allow mixing two Review Units in one PR. Stacked on the classification-only slice so this slice's own PR body is checked against a base that already recognizes the new file's path. This slice's tests also reference exact strings added in the first slice, so it lands third.

## Non-goals

- Does not change the existing missing-media or missing-animation checks — only adds a third condition after both pass.
- Does not touch `packages/**` product code.
- Does not add the rule to every skill in the repo — scoped to the four skills where this exact failure pattern was observed (`make-pr`, `visual-proof`, `land-stack`, `invoker-ops`).

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `node scripts/test-pr-body-validator.mjs` — passes, including new cases for a missing `Manually inspected:` line (fails), a present one (passes), and a bold-markdown/case-insensitive variant (passes).
- [x] `bash scripts/test-make-pr-skill.sh` — passes.
- [x] `bash scripts/test-visual-proof-skill.sh` — passes.
- [x] `bash scripts/test-land-stack-skill.sh` — passes.
- [x] `bash scripts/test-invoker-ops-skill.sh` — passes.
- [x] `bash scripts/test-prove-it-skill.sh` — passes.
- [x] `node scripts/test-review-unit-classification.mjs` — passes with `skills/prove-it/SKILL.md` added to the tooling-policy fixture list.

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

</details>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -13,7 +13,7 @@
 - Every step in a plan MUST be testable. Each implementation step must have a corresponding verification with a concrete, executable command that produces a clear pass/fail exit code (e.g. `pnpm test`, `git diff --name-only`). Do not use AI prompts for test tasks — use commands only.
 - Bug fix plans MUST follow a three-phase approach before any implementation:
   1. **Reproduce** -- Find or write a concrete reproduction case (a failing test or a command that demonstrates the bug). Report back the exact repro steps and observed vs. expected behavior. Do not proceed until the bug is reliably reproducible.
-  2. **Debug and report** -- Investigate and report: (a) the root cause — why the code is in the buggy state, and (b) the test gap — how the bug escaped existing tests (missing coverage, wrong assumptions, untested edge case, etc.).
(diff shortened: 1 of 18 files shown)


EdbertChan:

PR #8231, opened 2026-08-10
[Owner-Serve Crash Visibility](1) proof: bad poolMemberId dispatch doesn't crash the process

EdbertChan:
## Summary

A remote owner-serve process crashed silently after a queued task referenced a worktree poolMemberId with no matching config entry. No stack trace, no log line, nothing.

This adds a regression test that wires the real Orchestrator, TaskRunner, and LaunchDispatcher together with that exact broken shape and drives a dispatch through it.

It confirms `poll()` never throws and no unhandled rejection leaks. That specific dispatch path already fails only the one task, by design.

The actual crash mechanism was never conclusively pinned; it did not reproduce reliably under local stress testing. This test documents what is already proven safe.

## Review Claim

Approve that a task with a bad poolMemberId, dispatched through the real TaskRunner + LaunchDispatcher, fails only that one task and never crashes the owner process or leaks an unhandled rejection.

## Review Lane

proof

## Review Unit

proof

## Safety Invariant

Test-only file, `packages/app/src/__tests__/launch-dispatcher-bad-poolmember-crash-repro.test.ts`. No product code changes. Safe to review purely by reading the test and its assertions.

## Slice Rationale

Proof lands before the related hardening fix (2) so the reviewer sees what is already safe before reviewing what changes. This test is independently useful regardless of whether (2) lands: it is real coverage for a real incident, not scaffolding for the fix.

## Non-goals

Does not explain or fix the original silent crash -- that mechanism was never reproduced reliably enough to pin down. Does not touch any product code.

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `cd packages/app && pnpm test -- src/__tests__/launch-dispatcher-bad-poolmember-crash-repro.test.ts`
- [ ] `cd packages/app && pnpm test -- src/__tests__/launch-dispatcher.test.ts`

</details>

## Revert Plan

<details>
<summary>Revert Plan</summary>

- Safe to revert? Yes
- Revert command: `git revert <sha>`
- Post-revert steps: None
- Data migration? No

</details>

PR #8232, opened 2026-08-10
[Owner-Serve Crash Visibility](2) Log fatal headless startup errors synchronously before exit

EdbertChan:
## Summary

An owner-serve process on a remote host crashed with zero trace anywhere -- no stack, no stderr output, no invoker.log entry.

Both fatal-error paths in `main.ts` only wrote to `process.stderr` before calling `process.exit()`. That write is not guaranteed to flush before the process dies when stderr is a pipe.

Owner-serve's stderr is always a pipe: it runs under a supervisor, not an interactive terminal.

This routes both paths through the logger first. `FileAndDbLogger.error()` does a synchronous `appendFileSync`, so the error survives the exit regardless of stderr's flush timing.

This does not change how often the crash happens. It guarantees the next occurrence leaves a real stack trace instead of silence, which is what made the original incident undiagnosable.

## Review Claim

Approve that both fatal-error exit paths in `main.ts` (config load failure, and the top-level `runHeadless()` catch) now log through the synchronous file logger before `process.exit()`, instead of relying on a `process.stderr` write that can be lost.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

Purely additive logging on two already-fatal exit paths. Neither path returns or continues after logging -- both still call `process.exit()` immediately after, same as before. No change to what triggers the exit, only to what's recorded before it happens.

## Slice Rationale

This is a separate PR from (1) because they are different review claims: (1) proves one specific dispatch path is already safe, (2) changes runtime behavior on unrelated fatal-error paths. Ordered after (1) so the reviewer has the safety context first.

## Non-goals

Does not fix or explain the original crash's root cause -- it was never reproduced reliably enough to pin a specific mechanism, even under repeated local stress testing with the same broken config that triggered it live. Does not change retry behavior, exit codes, or add new recovery logic.

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `cd packages/app && pnpm run check:types` (repo root: `pnpm run check:types`)
- [ ] `cd packages/app && pnpm run build`
- [ ] `cd packages/app && pnpm test -- src/__tests__/launch-dispatcher-bad-poolmember-crash-repro.test.ts src/__tests__/launch-dispatcher.test.ts`

</details>

## Visual Proof

This diff has no UI-facing behavior; `main.ts` is flagged by path, not by anything user-visible changing. The proof is a real terminal run of the two exact functions this diff touches (`loadConfig`, `FileAndDbLogger`), captured in a single static frame.

Terminal output: loadConfig() throwing, labeled old vs. new catch-block behavior, then invoker.log going from absent to containing the logged fatal error

Manually inspected: I opened this image myself. It shows a real `vitest` run against the actual `loadConfig`/`FileAndDbLogger` source, not a mock. The old catch-block path leaves `invoker.log exists: false`; the new one (what this PR ships) leaves `invoker.log exists: true` with the real JSON log line printed, matching exactly what the patched catch block writes.

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

</details>

### DIFF
--- a/packages/app/src/__tests__/launch-dispatcher-bad-poolmember-crash-repro.test.ts
+++ b/packages/app/src/__tests__/launch-dispatcher-bad-poolmember-crash-repro.test.ts
@@ -1,0 +1,106 @@
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { SQLiteAdapter } from '@invoker/data-store';
+import { InMemoryBus } from '@invoker/test-kit';
+import { Orchestrator } from '@invoker/workflow-core';
+import { TaskRunner } from '@invoker/execution-engine';
+import { LaunchDispatcher } from '../launch-dispatcher.js';
+
+/**
+ * Repro for a live DO1 incident: a task's config pinned poolMemberId
+ * "local-worktree" against a pool member of that id, but worktreeTargets
+ * had no matching entry (the config didn't ship one). selectExecutor()
+ * throws for that task by design -- the question this test answers is
+ * whether that throw stays scoped to the one task/dispatch, or whether it
+ * escapes as an unhandled rejection and takes the whole owner process down.
+ */
+describe('LaunchDispatcher + real TaskRunner: bad poolMemberId does not crash the process', () => {
+  const adapters: SQLiteAdapter[] = [];
+
+  afterEach(() => {
+    for (const adapter of adapters.splice(0)) {
+      adapter.close();
+    }
+  });
+
+  it('fails only the offending dispatch instead of throwing out of poll()', async () => {
+    const persistence = await SQLiteAdapter.create(':memory:');
+    adapters.push(persistence);
+
+    const orchestrator = new Orchestrator({
+      persistence: persistence as any,
+      messageBus: new InMemoryBus(),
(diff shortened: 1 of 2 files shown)


EdbertChan:

| # | Pull Request | Link | |
|--:|---|---|---|
| 1 | (1) proof: real TaskRunner+LaunchDispatcher dispatch of a bad poolMemberId doesn't crash the process | [#8231](https://github.com/Neko-Catpital-Labs/Invoker/pull/8231) |  |
| 2 | (2) Log fatal headless startup errors synchronously before process.exit() | [#8232](https://github.com/Neko-Catpital-Labs/Invoker/pull/8232) | 👈 |

EdbertChan:

8232. #8232 — [Owner-Serve Crash Visibility](2) Log fatal headless startup errors synchronously before exit ← this PR
   Base: stack/EdbertChan/pr/owner-serve-crash-visibility/1-proof-real-taskrunner-launchdispatcher-dispatch--aeaa216f
   Link: https://github.com/Neko-Catpital-Labs/Invoker/pull/8232

PR #8241, opened 2026-08-10
Bugfix: headlessDeleteWorkflow throws on a benign FK race after the delete already succeeded

EdbertChan:
## Summary

A racing owner process can trip a foreign key error on an unrelated concurrent write while a workflow is being taken down.

The headless cancel path already tolerates this outcome when the workflow is confirmed gone.

The headless path for the other command lacked the same guard. It could report a foreign key failure and exit 1 even though the underlying row was already gone.

## Review Claim

Approve that the second headless command now reuses the same already-tested race check the first command uses, instead of throwing on a benign race outcome.

## Review Lane

behavior

## Review Unit

activation-surface

## Safety Invariant

Only widens an existing, already-tested check (previously private to the cancel path) to a second call site with the same race shape. Any other error still throws unchanged.

## Slice Rationale

Small, self-contained fix: one guard reused at one additional call site, with its own regression test in the same commit.

## Non-goals

Does not change the check's logic. Does not change behavior for any error that isn't this specific race.

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `cd packages/app && pnpm test -- src/__tests__/headless-delete-workflow-race.test.ts`

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Revert via: `git revert <sha>`
- Post-revert steps: None
- Data migration? No

</details>

### DIFF
--- a/packages/app/src/__tests__/headless-delete-workflow-race.test.ts
+++ b/packages/app/src/__tests__/headless-delete-workflow-race.test.ts
@@ -1,0 +1,78 @@
+/**
+ * Regression test for the delete-workflow race described in
+ * packages/data-store/src/__tests__/delete-workflow-race-fk-crash.test.ts.
+ *
+ * headlessDeleteWorkflow makes two calls that can race a concurrent delete:
+ *  1. preemptWorkflowExecution (cancel) — already guarded, see
+ *     headless-preempt-workflow-race.test.ts.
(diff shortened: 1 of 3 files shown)


EdbertChan:

EdbertChan:
@Mergifyio refresh

PR #8308, opened 2026-08-10
[Bugfix: Extract owner-capability write gate into named guard](1)[REFACTOR: Extract Function] Move the check into assertOwnerCapabilityForWritableOpen

EdbertChan:
## Summary

Only the owning process may open a database file for writing. Every other process gets a read-only connection or a temporary placeholder. That rule already existed and is unchanged.

This change only moves the existing check into its own named function, `assertOwnerCapabilityForWritableOpen`, inside `packages/data-store/src/sqlite-adapter.ts`. It runs at the same point, on the same condition, with the same error message.

Putting it in a named function makes it possible to test the rule on its own, without building a real database first.

## Review Claim

Approve moving the existing owner-capability check out of `SQLiteAdapter.create()` and into a named, exported function (`assertOwnerCapabilityForWritableOpen`), called from the same place with the exact same behavior, plus new tests that call the function directly.

## Review Lane

refactor

## Review Unit

ownership-refactor

## Safety Invariant

`SQLiteAdapter.create()` still throws the exact same error message, under the exact same condition (`isFile && requestWritable && !ownerCapability`), at the exact same point in the function. Only the condition now lives inside a named function instead of an inline `if` block.

## Slice Rationale

One slice: move the check into a named function and add tests that exercise it directly. Nothing else changes.

## Non-goals

No behavior change: the calling code in `packages/app/src/viewer-db-boundary.ts` is untouched, and the nearby exclusiveLocking and owner-pid checks next to this one are untouched. Only the inline check becomes a named function; the condition, error message, and throw point stay the same as before.

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `cd packages/data-store && pnpm test -- -t "assertOwnerCapabilityForWritableOpen"`
- [ ] `cd packages/app && pnpm test -- -t "rejects detached viewer writes when the real database exists"`

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

</details>

### DIFF
--- a/packages/data-store/src/__tests__/sqlite-adapter.test.ts
+++ b/packages/data-store/src/__tests__/sqlite-adapter.test.ts
@@ -2,7 +2,7 @@ import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import { mkdtempSync, rmSync, statSync, existsSync, readdirSync } from 'node:fs';
 import { join } from 'node:path';
 import { tmpdir } from 'node:os';
-import { SQLiteAdapter, isLaunchDispatchCandidateStale, runWithFreedStatement } from '../sqlite-adapter.js';
+import { SQLiteAdapter, isLaunchDispatchCandidateStale, runWithFreedStatement, assertOwnerCapabilityForWritableOpen } from '../sqlite-adapter.js';
 import type { Workflow, Conversation, WorkerActionWrite, TerminalSessionRecord, InAppPlanningSessionRecord } from '../adapter.js';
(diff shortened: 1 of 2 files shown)


EdbertChan:

8308. #8308 — [Bugfix: Extract owner-capability write gate into named guard](1)[REFACTOR: Extract Function] Move the check into assertOwnerCapabilityForWritableOpen ← this PR
   Base: plan/bugfix-extract-prepared-statement-free-pattern-into-named-helper
   Link: https://github.com/Neko-Catpital-Labs/Invoker/pull/8308

PR #8991, opened 2026-08-13
[Deploy self-kill via own kill-script](1a) Prove the kill-script matches its own ancestor

EdbertChan:
## Summary

`scripts/deploy-do1.sh` finds the old Invoker owner process to restart by walking `/proc` for processes whose working directory is the repo and whose argv text contains a match string.

The restart sequence itself runs as `setsid bash -c "<its own script text>"`. Its own working directory is also the repo, and its own argv text is that same script — containing the same match string.

The kill step matches its own ancestor and sends it SIGTERM, killing the whole restart sequence before it can restart the service.

This repro proves that mechanism in isolation, plus a second problem: the match string never matches the real owner process either, so a real old owner is never found even without the self-kill.

## Review Claim

The repro correctly demonstrates the self-kill mechanism (a buggy matcher kills its own ancestor process) and correctly demonstrates that a fixed matcher (excluding the caller's own process group) avoids it while still finding a real target.

## Review Lane

proof

## Review Unit

proof

## Safety Invariant

The repro is fully self-contained: it spawns its own throwaway processes under a temp directory and only ever signals processes it created itself. It does not touch any real Invoker process, and requires no live infrastructure. Safe to run locally or in CI.

## Slice Rationale

Proof lands before the fix (next slice) so the bug is demonstrated before the behavior change is reviewed, per this repo's review-compression ordering rule. The repro embeds the buggy and fixed matcher logic as inline, self-contained snippets rather than reading `scripts/deploy-do1.sh` itself, so it does not depend on that file's state and stays meaningful before and after the fix lands.

## Non-goals

Does not modify `scripts/deploy-do1.sh` itself — that's the next slice. Does not repro the *other*, already-fixed self-kill scenario in `repro-deploy-self-kill-aborts-restart.sh` (the owner's own shutdown cascade killing the deploy task) — that's a different mechanism, already guarded.

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `bash -n scripts/repro/repro-deploy-do1-kill-script-self-match.sh` (requires Linux `/proc`; run on a Linux host or in CI, not macOS)
- [ ] `bash scripts/repro/repro-deploy-do1-kill-script-self-match.sh`

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Revert command: `git revert d0e97df4c6`
- Post-revert steps: None
- Data migration? No

</details>

## Summary by CodeRabbit

* **Tests**
  * Added a standalone reproduction script for deployment restart process-matching issues.
  * Demonstrates the failure scenario where the restart sequence targets the wrong process.
  * Verifies the corrected behavior: the restart sequence avoids terminating itself and successfully stops the intended previous service.
  * Includes automated assertions that clearly report whether the reproduction passes or fails.

### DIFF
--- a/scripts/repro/repro-deploy-do1-kill-script-self-match.sh
+++ b/scripts/repro/repro-deploy-do1-kill-script-self-match.sh
@@ -1,0 +1,163 @@
+#!/usr/bin/env bash
+# Repro: deploy-do1.sh's python "find and kill the old owner process" step
+# matches and SIGTERMs its OWN ancestor process, aborting the restart
+# sequence before it ever reaches `systemctl restart slack-manager.service`.
+#
+# Root cause:
+#   The kill script (scripts/deploy-do1.sh, inside the `setsid bash -c '...'`
+#   restart block) finds kill targets by checking, for every /proc/<pid>:
+#     - realpath(cwd) == repo_root
+#     - the search substring appears in cmdline
+#   The `setsid bash -c '...'` wrapper that RUNS this whole restart sequence
+#   also `cd`s into repo_root, and its own cmdline is the literal script
+#   source text -- which necessarily contains the same search substring,
+#   since that substring is the pgrep/python match pattern written inline in
+#   the script itself. So the kill script matches its own ancestor and kills
+#   its own process group mid-sequence (observed live: RESTART_STATUS=143,
+#   i.e. SIGTERM, with no "restart"/"is-active" log lines ever emitted).
+#
+# Separately, the search substring itself ("packages/app/dist/main.js") is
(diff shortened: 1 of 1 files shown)


EdbertChan:

| # | Pull Request | Link | |
|--:|---|---|---|
| 1 | [Deploy self-kill via own kill-script](1a) Prove the kill-script matches its own ancestor | [#8991](https://github.com/Neko-Catpital-Labs/Invoker/pull/8991) | 👈 |
| 2 | [Deploy self-kill via own kill-script](1b) Exclude own process group from the owner matcher | [#8992](https://github.com/Neko-Catpital-Labs/Invoker/pull/8992) |  |

PR #9068, opened 2026-08-13
[Headless CLI bundling gap](1) Add CI-blocking functional smoke test for the headless CLI

EdbertChan:
## Summary

The headless CLI can crash on load without any visible error anywhere near the caller.

One workspace dependency's package.json pointed straight at raw, uncompiled source.

Left unbundled, Node tried to resolve that source file's own relative import at runtime and failed.

The crash happened during Electron's load phase as an uncaught exception, so the process never exited.

Every cron worker script that calls the headless CLI hung until its own timeout killed it.

In production this silently blocked a worker from filing any of its expected pull requests.

This PR adds a functional check that boots the real built CLI and asserts it doesn't crash.

The affected package is already fixed on master; this check exists as standing protection.

## Review Claim

Approve adding a functional boot check for the built headless CLI to required CI, as standing protection against this exact failure mode recurring with a different dependency.

## Review Lane

policy

## Review Unit

tooling-policy

## Safety Invariant

The added check only reads and executes the already-built artifact; it does not modify any build step, config, or product code, so it cannot change what ships.

## Slice Rationale

This is pure new verification tooling with no product-code change; it stands on its own regardless of any specific fix.

## Non-goals

Does not change any build configuration or product code.

Does not attempt to enumerate or fix every dependency that could exhibit this failure mode; that is a separate, already-passing concern.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `pnpm --filter @invoker/app build`
- [x] `bash scripts/test-suites/required/09-headless-cli-boots-smoke.sh`

</details>

## Revert Plan

<details>
<summary>Revert Plan</summary>

- Safe to revert? Yes
- Revert command: `git revert <sha>`
- Post-revert steps: None
- Data migration? No

</details>

PR #9069, opened 2026-08-13
[Headless CLI bundling gap](2) Dedupe noExternal and guard packages/app's build against the same gap recurring

EdbertChan:
## Summary

One workspace package's name appeared twice in the bundler's noExternal entries.

Three separate automated fixes had independently added and taken out that same entry.

None of them checked for an existing entry first, so the array ended up with a repeat.

The repeat is harmless to the bundler itself, just a sign of an unenforced invariant.

This PR collapses the repeat into one entry and adds a build-time check enforcing that invariant going forward.

The check fails the build if a bundled-source dependency is left out, or appears more than once.

Two dependencies are already known-safe to leave out and are recorded as explicit, reasoned exceptions.

## Review Claim

Approve collapsing the repeated entry into one and adding a build-time guard so the next omission or repeat fails the build instead of shipping silently.

## Review Lane

behavior

## Review Unit

activation-surface

## Safety Invariant

The array edit removes an inert duplicate string with no effect on what gets bundled; the new guard only runs at build time and cannot change what a successful build produces.

## Slice Rationale

The repeated entry and its guard are one claim: the guard is what makes the repeat provably a one-time fix rather than a recurring pattern.

## Non-goals

Does not change which packages are bundled versus left out; every currently-bundled dependency stays bundled.

Does not change `@invoker/cli`'s subprocess-spawn design or `@invoker/surfaces`'s separate packaging; both are recorded as deliberate, already-correct exceptions.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `pnpm --filter @invoker/app build`
- [x] `node packages/app/scripts/verify-noexternal-completeness.cjs`
- [x] `pnpm --filter @invoker/app test -- src/__tests__/workspace-import-resolution.test.ts`

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -400,6 +400,7 @@ jobs:
               bash scripts/test-suites/required/06-large-file-guardrail.sh
               bash scripts/test-suites/required/07-invalid-config-json.sh
               bash scripts/test-suites/required/08-build-artifact-surfaces-guard.sh
+              bash scripts/test-suites/required/09-headless-cli-boots-smoke.sh
               bash scripts/test-suites/required/10-dag-background-click-noop-guard.sh
               bash scripts/test-suites/required/15-owner-boundary-policy.sh
               bash scripts/test-suites/required/50-verify-executor-routing.sh
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -58,7 +58,7 @@
     }
   },
   "scripts": {
-    "build": "node scripts/verify-workspace-imports.cjs && tsup",
+    "build": "node scripts/verify-workspace-imports.cjs && node scripts/verify-noexternal-completeness.cjs && tsup",
     "dist": "electron-builder --publish never",
     "dist:linux": "electron-builder --linux AppImage deb --publish never",
     "dist:mac": "electron-builder --mac dmg --publish never",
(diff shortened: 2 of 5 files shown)


EdbertChan:

| # | Pull Request | Link | |
|--:|---|---|---|
| 1 | Add CI-blocking functional smoke test for the headless CLI | [#9068](https://github.com/Neko-Catpital-Labs/Invoker/pull/9068) |  |
| 2 | Dedupe noExternal and guard packages/app's build against the same gap recurring | [#9069](https://github.com/Neko-Catpital-Labs/Invoker/pull/9069) | 👈 |

EdbertChan:

9069. #9069 — [Headless CLI bundling gap](2) Dedupe noExternal and guard packages/app's build against the same gap recurring ← this PR
   Base: stack/EdbertChan/fix/slack-bug-scan-noexternal/add-ci-blocking-functional-smoke-test-headless-cli--113ac372
   Link: https://github.com/Neko-Catpital-Labs/Invoker/pull/9069

EdbertChan:
**packages/app/scripts/verify-noexternal-completeness.cjs:77 — Legitimate.** Lines 75-77 read only `depPackageJson.main` and derive `isRawTypeScript` solely from `mainEntry`; lines 79-80 then skip the dependency when that entry looks compiled, without examining `exports`, despite lines 3-6 explicitly promising `main/exports` coverage. A package can therefore pass this guard when `main` names compiled JavaScript but a runtime export condition names raw TypeScript; `packages/surfaces/package.json:6-10` also demonstrates why type-only and runtime export conditions must be distinguished.

PR #9309, opened 2026-08-16
Classify and repair disk-full SSH task failures via infra-repair

EdbertChan:
## Summary

A live SSH task failure captured tonight on a remote pool member at 100% disk usage showed git's own worktree/branch setup failing: "cannot lock ref ... unable to create directory." There was no free space to create the ref directory.

The failure classifier's pattern list didn't recognize that text, so infra-repair silently dropped every candidate with this shape, never attempting a repair. Two separate fix-workflow attempts hit this exact wall tonight before any agent work could start.

This adds a new failure class for it, and a repair handler that reclaims space on the target using the same routine disk-headroom already uses locally, then requeues the task.

## Review Claim

The reviewer is approving a new `ssh-disk-full` failure class plus an infra-repair handler that runs the existing remote disk-space routine and then requeues the task, using the exact error text captured from a real production incident tonight as the test fixture.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

The new handler only runs for a task whose error text matches the new disk-full signature; every other classification path is unchanged. Space reclamation reuses the same `cleanupRemoteInvokerHome` routine disk-headroom already runs in production, including its existing in-use-path protection.

## Slice Rationale

One behavior slice: the new classification plus its matching repair handler, proven against the exact error text a real incident produced tonight.

## Non-goals

Does not change any other failure classification. Does not touch disk-headroom's own worker (a separate, pre-existing gap where it isn't currently enabled on this pool is being tracked separately, not fixed here).

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `cd packages/workflow-graph && pnpm test -- src/__tests__/failure-classifier.test.ts`
- [ ] `cd packages/execution-engine && pnpm test -- src/__tests__/infra-repair-worker.test.ts`
- [ ] `npx tsc -p tsconfig.typecheck.json`

</details>

## Revert Plan

<details>
<summary>Revert Plan</summary>

- Safe to revert? Yes
- Revert via: `git revert 9a4be82e11`
- Post-revert steps: None
- Data migration? No

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of SSH infrastructure failures caused by full disks, including Git ref-lock errors.
  * Automatically cleans up remote invoker storage when disk capacity is exhausted.
  * Queues eligible tasks for retry after successful cleanup.
  * Records cleanup failures and cooldown skips for improved recovery tracking.
* **Tests**
[...]

PR #9312, opened 2026-08-16
Check final task status before treating a repair as crashed on infra

EdbertChan:
## Summary

A live PR #9309 bot-thread repair tonight hit an OAuth-expired error 4 times on early SSH attempts while credentials were still spreading across the pool, then self-healed and finished on its own.

Its full log still held that old error text, so the crash check flagged it as dead with no way to tell a passing hiccup apart from a real dead end. That wrongly told the planner to skip its own cooldown guard and fire a second, fully redundant repair while the first was still wrapping up.

Nothing was lost -- the second repair's own stale-head check caught the resulting race and refused to push -- but a whole second repair cycle ran for nothing.

## Review Claim

The reviewer is approving that the crash check now confirms the repair task's final status before trusting old error text in its log, using the exact real captured incident as the test fixture.

## Review Lane

policy

## Review Unit

tooling-policy

## Safety Invariant

A status-lookup failure still counts as a possible crash (unchanged fail-open direction), so this cannot silently hide a genuine dead task. Every other caller of this function is unaffected.

## Slice Rationale

One self-contained slice: a single function plus its test, proven against the exact incident data captured tonight.

## Non-goals

Does not touch the in-flight cooldown guard itself, only the signal that decides whether to bypass it. Does not change how repairs are submitted or how PR #9309 lands.

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `python3 scripts/test_mergify_admin_requeue_infra_signal.py`
- [ ] `python3 -m unittest discover -s scripts -p "test_mergify_admin_requeue*.py"`

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Revert via: `git revert 92f3ae7306`
- Post-revert steps: None
- Data migration? No

</details>

### DIFF
--- a/scripts/mergify_admin_requeue_infra_signal.py
+++ b/scripts/mergify_admin_requeue_infra_signal.py
@@ -29,12 +29,43 @@ def find_latest_workflow_id(plan_name: str, *, run_headless_fn=run_headless) -> 
     return matches[-1].get("id")
 
 
+def _repair_task_completed(workflow_id: str, *, run_headless_fn=run_headless) -> bool | None:
+    """True/False when the `repair` task's current status is known, None when
+    the status lookup itself failed (caller should fail open toward "crashed")."""
+    completed = run_headless_fn('headless_query query tasks "$2"', workflow_id)
+    if completed.returncode != 0:
+        return None
+    try:
+        tasks = json.loads(completed.stdout)
+    except (json.JSONDecodeError, TypeError):
+        return None
+    repair_task = next(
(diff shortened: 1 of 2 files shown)


EdbertChan:
@mergifyio queue

EdbertChan:
Mergify repair stopped: required check failed: UI Vitest. The retry cap was reached for current head 41f86ba6c499e6d9574278f7a9cc045f66ce8d0c.

PR #9372, opened 2026-08-16
Fix flaky-registry glob format and extend it to Vitest Workspace

EdbertChan:
## Summary

The flaky-test registry shipped tonight had a real gap: vitest resolves an exclusion path relative to each package's own folder, not the repo root. A registry entry written the natural way would silently exclude nothing.

This makes that mistake impossible to make going forward, and extends the same registry to the other vitest job in CI, gated so local test runs are never affected.

## Review Claim

The reviewer is approving a write-time guard on the registry's glob format, proven against the exact silent-failure this format caused, plus wiring the same registry into the Vitest Workspace job.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

The registry is empty today, so this ships with zero live behavior change. The new guard only rejects a glob shape at write time; it changes nothing about how an already-valid glob is applied. The Vitest Workspace wiring only activates when `CI` is set, so local runs are unaffected either way.

## Slice Rationale

One slice: fix the format gap in the mechanism just shipped, and extend it to the one other vitest job while the wiring pattern is fresh, proven end-to-end with a stubbed run rather than the real multi-minute suite.

## Non-goals

Does not touch Playwright, which needs a different mechanism entirely (its shards are hardcoded file lists in ci.yml, not a CLI flag) — scoped separately.

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `node scripts/test-flaky-test-registry.mjs`
- [ ] `bash scripts/test-workspace-test-flaky-wiring.sh`
- [ ] `node scripts/test-ci-workflow-merge-queue-policy.mjs`

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Revert via: `git revert ab2d9f05b0`
- Post-revert steps: None
- Data migration? No

</details>

### DIFF
--- a/scripts/flaky-test-registry.mjs
+++ b/scripts/flaky-test-registry.mjs
@@ -27,6 +27,15 @@ export function readRegistry(registryText) {
   return parsed;
 }
 
+// vitest resolves --exclude relative to each package's own directory (the
+// cwd it runs in, whether invoked via `pnpm --filter <pkg> test` or `pnpm -r
+// test`), never the repo root. A repo-root-relative glob like
+// "packages/ui/src/foo.test.ts" silently matches nothing and quarantines
+// nothing -- confirmed live: it left the target test running. Requiring a
+// double-star wildcard prefix on vitest-file targets keeps every such entry
+// package-relative-safe by construction instead of relying on whoever adds
+// an entry to remember this. Suite targets are shell script relpaths, not
+// vitest globs, so the requirement does not apply to them.
 export function quarantineInRegistry(registry, target, { reason, source, now, kind }) {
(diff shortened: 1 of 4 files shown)


EdbertChan:
### Revision history

| 1 | initial | `ab2d9f0` |  | 2026-08-16 05:44 UTC |
| 2 | content | [`ab2d9f0 → 603be57`](https://github.com/Neko-Catpital-Labs/Invoker/compare/ab2d9f05b03614133ef67a81a2a4022d63a2ed00...603be570eb07668e43d42714ea280236be15179f) |  | 2026-08-16 05:44 UTC |

PR #9467, opened 2026-08-16
[Electron Interrupted Extraction](1) Reproduce: interrupted Electron extraction leaves a half-written dist/

EdbertChan:
## Summary

Every playwright CI job and several others have been failing at fixture setup, before any test-specific code runs, on the self-hosted runner fleet.

Live evidence on all 5 of those runners: the Electron package's platform binary is missing, while the archive it comes from is present and valid. Something interrupts extraction partway through.

This PR adds a case reproducing exactly that: a killed extraction that writes some files but not the platform binary.

## Review Claim

The reviewer is approving a new case in the existing Electron provisioning repro script that demonstrates today's real failure mode, confirmed failing against the current, unfixed extraction code.

## Review Lane

proof

## Review Unit

proof

## Safety Invariant

This PR only adds a repro case; it does not change any product code. Confirmed the new case fails against the current, unmodified extraction path (see Test Plan) — this is a genuine before-state, not a case written to already pass.

## Slice Rationale

Split from the fix itself (paired PR, landing separately) so the failure is demonstrated before the change that addresses it is approved, per this repo's proof-before-fix convention. Kept separate from the other CI-failure fixes landing in this same push, since this is an independent root cause behind a different set of failing jobs.

## Non-goals

Does not fix the extraction bug this demonstrates — see the paired fix PR. Does not address a separate, pre-existing gap in this same script's original case (its fake hardcodes the Linux binary name, so it only runs reliably under `npm_config_platform=linux` on a non-Linux dev machine) — flagged, not fixed here.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `npm_config_platform=linux bash scripts/repro/repro-mece-04-remote-electron-provisioning.sh` against this PR's code alone (current unfixed `scripts/electron.cjs`) — exit 1, `repro: an interrupted extraction left a half-written dist/ (has locales/ but not the platform binary) -- this is the corrupted state observed on the real CI fleet`

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

</details>

### DIFF
--- a/scripts/repro/repro-mece-04-remote-electron-provisioning.sh
+++ b/scripts/repro/repro-mece-04-remote-electron-provisioning.sh
@@ -204,3 +204,68 @@ if [[ ! -x "$TMP_DIR/repo/node_modules/electron/dist/electron" ]]; then
 fi
 
 echo "remote-electron-provisioning fixed: fallback extraction succeeds without a host unzip executable"
+
+# Regression: a self-hosted CI runner reuses the same on-disk workspace across
+# job runs. If a run is killed (job cancellation, timeout-minutes) while
+# extract-zip is mid-write, an extraction straight into the live dist/
+# directory can leave some files present but not the platform binary --
+# resolveInstalledElectronBinary()'s existence-only check must never treat
+# that half-written state as installed, and a subsequent install-only must
+# never leave dist/ in that half-written shape either.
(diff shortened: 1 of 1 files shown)


EdbertChan:
Mergify repair stopped: required check failed: UI Vitest. The retry cap was reached for current head 581f50f71805a7b8b5a17333f0ad62994afd7a17.

PR #11137, opened 2026-08-28
[Python comment policy](1) Grandfather pre-existing files, document check:comments

EdbertChan:
## Summary

The Added Comment Policy check already covers Python, merged yesterday. It only looks at diff lines, so an ordinary edit was never at risk of flagging old comments.

A whole-file reformat is different: git shows every line as a fresh pair, comments included. That would force stripping legacy comments as a side effect of unrelated work.

This grandfathers the files that already carried comments before Python coverage landed. Every other Python file gets full enforcement immediately.

It also adds a note about `pnpm run check:comments` to the PR-authoring skill, since nothing mentioned this gate before now.

## Review Claim

Approve grandfathering the specific Python files that already had comments before this check covered Python, and noting the check in the PR-authoring skill.

## Review Lane

policy

## Review Unit

tooling-policy

## Safety Invariant

The grandfather list is an exact-path match against a fixed, closed set of files; it cannot silently grow to cover a new file. Any Python file not on the list, including one added tomorrow, is fully enforced from its first line.

## Slice Rationale

The exemption and its skill note describe one policy decision; keeping them apart would leave the note pointing at behavior nobody could verify.

## Non-goals

Does not touch the actual Python comment-scanning logic from #10693; that already works correctly.

Does not retroactively require cleaning up the grandfathered files' existing comments; that is a separate, optional future pass.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `node scripts/test-check-added-comments.mjs` — new fixtures: a genuinely new comment in a grandfathered file is not flagged; the identical comment in a same-named-but-different file is flagged.
- [x] Fail-before/pass-after: ran the un-fixed `origin/master` checker against the grandfathered-file fixture directly — 1 violation before this change, 0 after.
- [x] `pnpm run check:comments` — passes clean against this PR's own diff.

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Revert command: `git revert a4cc9dd`
- Post-revert steps: None
- Data migration? No

</details>


EdbertChan:

11137. #11137 — [Python comment policy](1) Grandfather pre-existing files, document check:comments ← this PR
   Base: master
   Link: https://github.com/Neko-Catpital-Labs/Invoker/pull/11137

PR #11214, opened 2026-08-28
[Production owner service profile detection](1) Recognize the production owner service as production

EdbertChan:
## Summary

Every conflict repair and failed-check repair on the deployed host was silently no-op-ing, even after last night's related fix landed and today's build gap closed.

The real cause: the client-side code that resolves which socket to talk to has no way to know it's running as a child of the real production service. It always assumes it might be a developer's incidental local git checkout, and redirects to an isolated dev socket that nothing is listening on.

The production service itself already carries a marker proving it's the real thing. This slice makes the client-side resolver check for that marker before falling back to dev-profile detection.

## Review Claim

`resolveActiveInvokerProfileEnv` now returns the plain production environment immediately when it detects it is running as (or under) the production owner service, instead of always spawning the dev-profile detection script.

## Review Lane

behavior

## Review Unit

contract

## Safety Invariant

This only short-circuits to the production shape when a specific environment marker is present, and that marker is only ever set by the production owner service itself and inherited by its own child processes. A normal developer checkout without that marker set falls through to the exact same dev-profile detection as before, unchanged.

## Slice Rationale

One targeted fix isolated from the two unrelated things it was masking (a stale build artifact, an unbuilt package) — both already shipped as their own separate slices.

## Non-goals

Does not change `scripts/with-invoker-development-profile.mjs` itself, does not change anything about how the production owner service starts, and does not touch any repair-worker logic beyond the socket it now correctly reaches.

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `pnpm --filter @invoker/contracts test`

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Revert command: `git revert 389692023d`
- Post-revert steps: None
- Data migration? No

</details>

### DIFF
--- a/packages/contracts/src/__tests__/resolve-active-profile-env.test.ts
+++ b/packages/contracts/src/__tests__/resolve-active-profile-env.test.ts
@@ -2,9 +2,9 @@ import { spawnSync } from 'node:child_process';
 import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
-import { afterEach, describe, expect, it } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
 
-import { resolveActiveInvokerProfileEnv, resolveRepoRoot } from '../index.js';
+import { resolveActiveInvokerProfileEnv, resolveInvokerIpcSocketPath, resolveRepoRoot } from '../index.js';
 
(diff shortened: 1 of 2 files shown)


EdbertChan:

11214. #11214 — [Production owner service profile detection](1) Recognize the production owner service as production ← this PR
   Base: master
   Link: https://github.com/Neko-Catpital-Labs/Invoker/pull/11214

PR #11942, opened 2026-09-04
[Codex OAuth Race](1) Widen the Codex OAuth proactive-refresh window from 5 minutes to a day

EdbertChan:
## Summary

Five machines share one Codex login, copied as a single file. Its refresh token works once: whichever machine rotates it first invalidates every other copy.

A five-minute window decided when rotation should happen, checked hourly by the one machine meant to own it. Five other machines, constantly busy, checked the same boundary far more often on their own.

That narrow window was almost always won by a busy machine instead of the intended owner. The loser then failed outright next time it needed its now-dead copy.

## Review Claim

Approve widening that one window from five minutes to a full day, so the intended owner's much slower check gets dozens of chances to win before a busy machine's own check ever gets close enough to race it.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

The change is one constant, used in exactly one staleness decision that only ever governs Codex credentials. The equivalent decision for the separate Claude credential path is untouched, still keyed to its own five-minute constant. The existing seven-day fallback staleness check, which serves a different purpose (catching a copy that was never properly stamped at all), is also untouched.

## Slice Rationale

One constant, one decision function, one root cause — nothing here depends on or blocks any other change.

## Non-goals

Does not change the Claude credential refresh path, the seven-day fallback check, how often the owning machine's check itself runs, or anything about how a machine obtains its very first login.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `bash node_modules/.bin/vitest run src/__tests__/codex-oauth-refresh.test.ts -t "closing the narrow multi-host refresh-collision window"` (run from `packages/execution-engine`) — fails with `expected false to be true` before the change, passes after
- [x] `bash node_modules/.bin/vitest run src/__tests__/codex-oauth-refresh.test.ts src/__tests__/claude-oauth-refresh-worker.test.ts src/__tests__/claude-oauth-refresh.test.ts` (run from `packages/execution-engine`) — 52/52 pass, no regressions
- [x] `pnpm run check:comments` — clean

This is a live-timing/live-OAuth-behavior bug: the actual collision (two real machines racing an external OAuth token endpoint) cannot be reproduced deterministically in a unit test. The test above proves the fixed decision function's new boundary directly; the live effect was observed and diagnosed against a real incident on the production fleet (see below), not reproduced synthetically.

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

</details>

---

**Evidence behind this fix**, gathered from the live production fleet before writing any code:

- A real task's `execution.error` field showed the exact failure: `codex_login::auth::manager: Failed to refresh token: Your access token could not be refreshed because your refresh token was already used.` — on `remote_digital_ocean_3`, 2026-09-04T00:55:55Z.
- `invoker-cli query workflows` confirmed 27 separate CI-repair workflows failed with the identical symptom in the same ~9-minute window (00:49–00:58 UTC), consistent with one shared-credential incident, not 27 independent causes.
- Read the current production `~/.codex/auth.json`'s access-token JWT directly: real lifetime is 10 days (`exp - iat = 864000` seconds).
- Ran a real `codex exec` call against a token with ~10 days of remaining life: the auth file was byte-for-byte unchanged afterward, confirming Codex's own client does not refresh far ahead of expiry — only close to its own boundary, which is what makes the shared five-minute window the actual collision point.
- Read `codex-oauth-refresh.ts` and `claude-oauth-refresh-worker.ts` in full: confirmed the one machine meant to own redistribution only checks hourly, while the other five machines' own Codex clients evaluate the identical boundary on every task run, with no coordination between any of these checks.

### DIFF
--- a/packages/execution-engine/src/__tests__/claude-oauth-refresh-worker.test.ts
+++ b/packages/execution-engine/src/__tests__/claude-oauth-refresh-worker.test.ts
@@ -233,7 +233,7 @@ function jwtWithExp(expSeconds: number): string {
 }
 
 function codexAuthJson(now: number, overrides: { lastRefresh?: string; expSeconds?: number } = {}): string {
-  const expSeconds = overrides.expSeconds ?? Math.floor((now + 60 * 60 * 1000) / 1000);
+  const expSeconds = overrides.expSeconds ?? Math.floor((now + 30 * 60 * 60 * 1000) / 1000);
   return JSON.stringify({
     auth_mode: 'chatgpt',
     tokens: {
--- a/packages/execution-engine/src/__tests__/codex-oauth-refresh.test.ts
+++ b/packages/execution-engine/src/__tests__/codex-oauth-refresh.test.ts
@@ -1,8 +1,8 @@
 import { describe, expect, it } from 'vitest';
 
-import { OAUTH_EXPIRY_BUFFER_MS } from '../claude-oauth-refresh.js';
 import {
   applyRefreshedCodexToken,
+  CODEX_JWT_EXPIRY_BUFFER_MS,
   CODEX_LAST_REFRESH_MAX_AGE_MS,
   isCodexAuthExpiring,
   parseCodexAuthFile,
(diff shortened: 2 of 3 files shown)


EdbertChan:

11942. #11942 — [Codex OAuth Race](1) Widen the Codex OAuth proactive-refresh window from 5 minutes to a day ← this PR
   Base: master
   Link: https://github.com/Neko-Catpital-Labs/Invoker/pull/11942

PR #11945, opened 2026-09-04
[CLI SEA Build Prefix Strip](1) Fix esbuild stripping node: from require("sea") in the CLI SEA build

EdbertChan:
## Summary

esbuild, the bundler our command-line package's self-contained-binary build runs on, quietly drops the `node:` prefix from one specific Node runtime import while packaging it.

Every freshly packaged binary then crashed the instant it started on an otherwise healthy machine, since that one newer import has no unprefixed alias, unlike the older ones esbuild also touches.

A controlled side-by-side build, isolated from the real program entirely, narrowed this to esbuild's own prefix-dropping behavior, ruling out this repo's own source and every installed dependency along the way.

## Review Claim

Approve rewriting that one mangled reference back to its correct form immediately after packaging finishes, before the file is sealed into its final self-contained shape.

## Review Lane

policy

## Review Unit

tooling-policy

## Safety Invariant

The rewrite only ever touches one exact reference, byte for byte, inside a file that already exists purely as packaging output -- nothing upstream of it changes, and an already-correct reference is left completely alone.

## Slice Rationale

One packaging step, one narrow correction, kept apart from the earlier slice that made the packaged program aware of a workflow's own removal in the first place.

## Non-goals

Does not change anything about how the program itself behaves once it starts. Does not touch any other packaged reference; only the one exact string this repo is currently known to hit.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `node scripts/patch-sea-builtin-prefix.test.mjs` -- 4/4 pass, covering the rewrite, an already-correct reference staying untouched, and unrelated content staying untouched
- [x] `INVOKER_TARGET_PLATFORM=linux INVOKER_TARGET_ARCH=x64 node scripts/build-cli-standalone.mjs` -- succeeds end to end
- [x] `grep -ac 'require("sea")' <built binary>` -- `0`
- [x] `grep -ac 'require("node:sea")' <built binary>` -- `1`
- [x] Copied the built binary to a real Linux machine and ran it there directly: `--version` prints the current version, `--help` lists the workflow-removal command added in the prior slice, no crash

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None
- Data migration? No

</details>

---

**Evidence behind the root cause**, gathered before writing any fix:

- Grepped this repo's own source and every installed package under `node_modules` for the exact broken reference text -- zero matches anywhere, ruling out a first-party bug or a vendored copy carrying it.
- Built a minimal, isolated repro with the packaging tool alone (no application code at all): a one-line file importing the correctly-prefixed Node built-in, packaged with the identical flags this repo uses, reproduced the exact same mangled output.
- Ran the same minimal repro a second way, with the built-in named explicitly rather than left to automatic detection: identical mangled output either way, ruling out an automatic-detection-only explanation.
- Applied the correction to the minimal repro's output by hand and ran it directly: it resolved and ran correctly, confirming the correction itself is sufficient before applying it anywhere real.
- Rebuilt the real, complete packaged program with the fix applied, copied the result to a real target machine, and ran it there directly -- it started, reported the right version, and listed the command added in the prior slice.

---

> [!NOTE]
> **Low Risk**
> Narrow post-bundle string fix on packaging artifacts only; no runtime behavior change beyond fixing broken standalone CLI startup.
> 
> **Overview**
> Fixes **standalone CLI binaries crashing on startup** when esbuild emits bare `require("sea")` instead of `require("node:sea")` — `sea` has no unprefixed Node alias, unlike older built-ins.
> 
> Adds a **post-bundle packaging step** in `build-cli-standalone.mjs` that rewrites only those bare `require("sea")` / `require('sea')` calls in the tsup output before the SEA blob is generated. The logic lives in `scripts/lib/patch-sea-builtin-prefix.mjs` with a small allowlist (`sea` only); already-correct and unrelated requires are left alone.
> 
> **Tests:** `patch-sea-builtin-prefix.test.mjs` covers rewrite, idempotency, and non-matches; the root `test` script now runs it in CI.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d053e3ff6ef9bb0e2f068970172458c05e788a8e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed standalone CLI builds so Node.js built-in modules are referenced correctly when packaged, improving reliability during startup and execution.

- **Tests**
  - Added automated coverage for standalone CLI packaging behavior, including quoted imports, already-correct references, and unrelated module imports.
  - Included the new validation in the standard test suite.


EdbertChan:

11945. #11945 — [CLI SEA Build Prefix Strip](1) Fix esbuild stripping node: from require("sea") in the CLI SEA build ← this PR
   Base: master
   Link: https://github.com/Neko-Catpital-Labs/Invoker/pull/11945

EdbertChan:
Invoker repair dispatch was not acknowledged. The request was recorded as pending, then closed without consuming the code-repair retry budget; automatic retry remains enabled for current head aeaf0ff.

PR #12040, opened 2026-09-09
[Flaky Vitest Workspace](1) Measure materialized bytes instead of heap delta

EdbertChan:
## Summary

The materialization proof measured the JS heap. It read `process.memoryUsage().heapUsed` before and after `listWorkflows()` and asserted the delta stayed under 1MB, inside an `it.fails` block.

`it.fails` inverts the outcome: the block is green only while that assertion throws. So the whole suite depended on garbage collection *not* running at a convenient moment.

When GC did land the delta under 1MB the assertion stopped throwing, `it.fails` went red, and `required-fast / Vitest Workspace` went red with it.

A repro script measured both instruments over 12 identical runs. The heap delta swung 33,350,032 bytes and fell under the ceiling on 8 of 12 runs — it even goes negative, because GC reclaims more than the query allocates.

Byte-length of the serialized rows measures the same claim without asking the heap: 4,091,671 bytes every run, spread of zero.

## Review Claim

Approve measuring the materialized payload by serialized byte-length instead of by heap delta, keeping the same `it.fails` claim that the query returns unbounded results.

## Review Lane

proof

## Review Unit

proof

## Safety Invariant

One test file, no product code. The `it.fails` wrapper, the workflow count, and the 1,000,000 ceiling are all unchanged, so the claim under proof is identical — only the instrument changes, from a nondeterministic heap reading to a deterministic byte count of the value the assertion is already about. The measured value is 4.09x the ceiling and has a spread of zero across runs, so no plausible environment flips it.

## Slice Rationale

One test file on its own, separate from the worker behaviour change in slice 2. They share a cause and nothing else.

## Non-goals

Does not change `listWorkflows`, `listWorkflowsPaged`, or any product code.

Does not change what the proof asserts, only how it measures.

Does not touch the sibling `loadWorkflowTaskSnapshot` proof, which already uses a deterministic length assertion.

## Correction

An earlier revision of this body cited `2,441,671` bytes. That figure came from a hand-rolled model of the rows rather than from `SQLiteAdapter.listWorkflows()`, and was wrong. The repro script measures the real adapter output: `4,091,671` bytes.

## Prior art

This is `02643eec`, authored 2026-09-04 during the incident and never merged. A second never-merged fix for the same flake exists at `5bd9a627`, which also wraps the inserts in a transaction. `02643eec` is the narrower of the two and is landed here as-is rather than authoring a third version.

Consolidation signal: two prior independent commits (`02643eec`, `5bd9a627`) already fixed this exact assertion and neither reached master. Both were written on disposable CI-repair branches. Worth a follow-up on why repair branches are not surfaced for landing; not performed here.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] Fail-before, the exact failure mode. Raising the ceiling to `1_000_000_000_000` simulates GC landing the delta under it, and reproduces the red state seen in CI:

  ```
  FAIL  src/__tests__/unbounded-workflow-select.repro.test.ts > listWorkflows materializes every row into JS objects
  Error: Expect test to fail
   Test Files  1 failed (1)
        Tests  1 failed | 4 passed (5)
  ```

- [x] Pass-after, three consecutive runs of `bash node_modules/.bin/vitest run src/__tests__/unbounded-workflow-select.repro.test.ts` from `packages/data-store` — `Tests  5 passed (5)` each time
- [x] Repro script, committed as `repro/08-flaky-materialization-proof-heap-instrument.mjs`. Run `pnpm --filter @invoker/data-store build` then `node repro/08-flaky-materialization-proof-heap-instrument.mjs`. It measures both instruments over 12 identical runs and exits non-zero if the new one ever varies or dips under the ceiling:

  ```
  heapUsed delta (old)   min= -11,977,616  max=  21,203,688  spread=  33,181,304  runs_under_1,000,000=8/12
  serialized bytes (new) min=   4,091,671  max=   4,091,671  spread=           0  runs_under_1,000,000=0/12

  PASS: serialized-byte instrument is identical on every run (4,091,671 bytes)
  PASS: serialized-byte instrument never fell under the threshold, so the block cannot flake
  OBSERVED: heap instrument varied by 33,181,304 bytes across identical runs (8 run(s) under the threshold)
  ```

  The old instrument fails the ceiling on 8 of 12 runs, so this was not a rare flake; the new one has a spread of zero
- [x] `pnpm run check:types` — exit 0
- [x] `pnpm run check:comments` — exit 0

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None; reverting restores the flaky heap measurement
- Data migration? No

</details>

### DIFF
--- a/packages/data-store/src/__tests__/unbounded-workflow-select.repro.test.ts
+++ b/packages/data-store/src/__tests__/unbounded-workflow-select.repro.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { Buffer } from 'node:buffer';
 import { mkdtempSync, rmSync } from 'node:fs';
 import { join } from 'node:path';
 import { tmpdir } from 'node:os';
@@ -54,24 +55,23 @@ describe('unbounded workflow SELECT (ui-read-scale proof)', () => {
 
   it.fails('listWorkflows materializes every row into JS objects', () => {
     const workflowCount = 5_000;
+    const descriptionPayload = 'x'.repeat(256);
     for (let i = 0; i < workflowCount; i++) {
       adapter.saveWorkflow({
         id: `wf-${i}`,
         name: `Workflow ${i} with a longer name to increase memory footprint`,
-        description: `Description for workflow ${i} that adds more bytes per row`,
+        description: `Description for workflow ${i}: ${descriptionPayload}`,
         status: 'pending',
         createdAt: new Date().toISOString(),
         updatedAt: new Date().toISOString(),
       });
     }
 
-    const before = process.memoryUsage().heapUsed;
     const workflows = adapter.listWorkflows();
-    const after = process.memoryUsage().heapUsed;
-    const memoryDelta = after - before;
+    const materializedBytes = Buffer.byteLength(JSON.stringify(workflows));
 
     expect(workflows).toHaveLength(workflowCount);
-    expect(memoryDelta).toBeLessThan(1_000_000);
+    expect(materializedBytes).toBeLessThan(1_000_000);
   });
 
   it('listWorkflowsPaged returns bounded results with pagination metadata', () => {
--- a/repro/08-flaky-materialization-proof-heap-instrument.mjs
+++ b/repro/08-flaky-materialization-proof-heap-instrument.mjs
@@ -1,0 +1,100 @@
+#!/usr/bin/env node
+import { mkdtempSync, rmSync } from 'node:fs';
+import { join } from 'node:path';
+import { tmpdir } from 'node:os';
+import { Buffer } from 'node:buffer';
+
+let SQLiteAdapter;
+try {
+  ({ SQLiteAdapter } = await import('../packages/data-store/dist/index.js'));
+} catch (error) {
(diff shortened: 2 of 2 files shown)


EdbertChan:
This pull request is part of a stack:
1. [Flaky Vitest Workspace](1) Measure materialized bytes instead of heap delta ([#12040](https://github.com/Neko-Catpital-Labs/Invoker/pull/12040)) 👈
1. [Flaky Vitest Workspace](2) Chain babysit investigations instead of firing them in parallel ([#12041](https://github.com/Neko-Catpital-Labs/Invoker/pull/12041))

EdbertChan:

12040. #12040 — [Flaky Vitest Workspace](1) Measure materialized bytes instead of heap delta ← this PR
   Base: master
   Link: https://github.com/Neko-Catpital-Labs/Invoker/pull/12040

EdbertChan:
Invoker Mergify babysitting is paused: this is the current bottom PR in the stack, but it is missing the `admin-bypass` label. Please tag this PR with `admin-bypass` before babysitting can continue.

EdbertChan:
Invoker Mergify babysitting is paused: this is the current bottom PR in the stack, but it is missing the `admin-bypass` label. Please tag this PR with `admin-bypass` before babysitting can continue.

PR #12178, opened 2026-09-12
Bugfix: a failed codex model probe is indistinguishable from an empty one

EdbertChan:
## Summary

The Codex model probe now distinguishes a failed probe from a successful probe that found no models.

The caller can therefore preserve graceful fallback behavior without treating a failed capability query as an empty result.

## Review Claim

The model-probe helper returns a discriminated result, so a failed probe is no longer indistinguishable from a successful empty probe.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

When the probe succeeds, the entries surfaced are byte-for-byte what they are today; only the failure path changes.

## Slice Rationale

This slice contains the probe return shape and its single caller. The separate proof-only workflow gate was empty and is not published as a PR.

## Non-goals

No timeout or cache changes. No operator-facing wording changes. No expansion of the hardcoded fallback. No edits to the Claude or OMP agents.

### Before

```mermaid
graph TD
    A["Codex probe"] --> B["Array result"]
    B --> C["Caller treats empty as fallback"]
```

### After

```mermaid
graph TD
    A["Codex probe"] --> B["Discriminated result"]
    B --> C["Caller branches on probe outcome"]
```

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `pnpm --filter @invoker/execution-engine exec vitest run src/__tests__/codex-model-discovery-failure.test.ts`
- [x] `bash scripts/scrub-handoff-artifacts.sh`

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Revert command: `git revert <published-commit-sha>`
- Post-revert steps: Re-run the focused execution-engine test.
- Data migration? No

</details>

### DIFF
--- a/packages/execution-engine/src/__tests__/codex-model-discovery-failure.test.ts
+++ b/packages/execution-engine/src/__tests__/codex-model-discovery-failure.test.ts
@@ -1,0 +1,57 @@
+import { spawnSync } from 'node:child_process';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import type { ExecutionModelOption } from '../agent.js';
+import { CodexExecutionAgent } from '../agents/codex-execution-agent.js';
+
+vi.mock('node:child_process', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('node:child_process')>();
+  return {
+    ...actual,
+    spawnSync: vi.fn(),
+  };
+});
+
+type ModelDiscoveryResult =
+  | { kind: 'success'; models: readonly ExecutionModelOption[] }
+  | { kind: 'failed' };
+
+type ModelDiscoveryTestAccess = {
+  discoverSupportedModels(): ModelDiscoveryResult;
+};
+
+function discoverSupportedModels(agent: CodexExecutionAgent): ModelDiscoveryResult {
+  return (agent as unknown as ModelDiscoveryTestAccess).discoverSupportedModels();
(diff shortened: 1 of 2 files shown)


EdbertChan:
Reviewer note from the plan author — one behavior change goes slightly beyond the safety invariant as written.

The invariant said: *"when the probe succeeds, the entries surfaced are byte-for-byte what they are today; only the failure path changes."* That is true for a probe that succeeds **with** entries, and for a failed probe. It is not quite true for a probe that succeeds with **zero** entries.

Before: a successful-but-empty probe fell through `discovered.length > 0` and used `CODEX_FALLBACK_MODELS`.
After: it is `{ kind: 'success', models: [] }` and the empty set is preserved — asserted directly by the new spec:

```
it('represents a successful probe with zero entries distinctly and preserves the empty result')
  expect(agent.supportedModels).toEqual([]);
```

I think that is the correct semantics — separating "codex told us nothing" from "we could not ask" is the entire point of the change, and silently substituting a stale list for a genuine empty answer is the same class of bug being fixed here. But it is a real change in a second case, it was not named in the invariant, and it deserves a decision rather than a nod: if codex ever returns an empty model list, dispatch will now reject every model instead of falling back.

Flagging it rather than letting it ride, since the invariant is what a reviewer would otherwise scope their reading to.

PR #12847, opened 2026-09-17
[Bazel pilots](4) Give Bazel pilot tests the pnpm test job's environment

EdbertChan:
## Summary

The Bazel cache pilot now gives tests the same environment as the normal pnpm test job.

Bazel drops HOME, CI and the git config from tests, and the pilot skipped the git identity and build-artifact steps. Git-backed tests failed with Committer identity unknown.

The remaining cache-pilot failures also fail under plain pnpm on master.

## Review Claim

Approve passing HOME, CI, GITHUB_ACTIONS and GIT_CONFIG_GLOBAL to Bazel tests and mirroring the pnpm job's setup steps.

## Review Lane

policy

## Review Unit

tooling-policy

## Safety Invariant

Both Bazel pilots stay outside Mergify merge conditions, which scripts/test-bazel-overlay.mjs asserts, so this change cannot block or unblock any merge.

## Slice Rationale

The environment is the last pilot stage; its effect is only visible once the earlier slices let Bazel run.

## Non-goals

Fixing the 52 app, cli, and execution-engine tests that fail under plain pnpm on master too.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `node scripts/test-bazel-overlay.mjs`
- [x] `node scripts/test-ci-workflow-merge-queue-policy.mjs`
- [x] Run 35165180896 (before): 87 failing tests, including `Committer identity unknown`
- [x] Run 35167176948 on this stack's tip: 19 of 22 packages pass; all 52 failing tests also fail under plain `pnpm --filter <pkg> test` on the same runner (run 35165944636)

</details>

## Revert Plan

<details>
<summary>Revert Plan</summary>

- Safe to revert? Yes
- Revert command: `git revert <sha>`
- Post-revert steps: None
- Data migration? No

</details>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12844, opened 2026-09-17
[Bazel pilots](1) Point the overlay self-check at the step that runs Bazel

EdbertChan:
## Summary

The Bazel overlay self-check ran the wrong CI step, so bazel-cache-pilot and quality / Bazel Overlay failed before Bazel ran.

PR #12672 added a credentials step that also carries the BuildBuddy key. The check picked it first, never saw bazelisk, and overwrote .bazelrc.user with a dummy key.

The check now picks the step that actually runs bazelisk.

## Review Claim

Approve that the overlay self-check tests the pilot step that invokes bazelisk.

## Review Lane

policy

## Review Unit

tooling-policy

## Safety Invariant

Both Bazel pilots stay outside Mergify merge conditions, which scripts/test-bazel-overlay.mjs asserts, so this change cannot block or unblock any merge.

## Slice Rationale

This is the first failing stage of both pilots; later slices fix the stages it hid.

## Non-goals

No change to what the pilots run. The image, RBE targets, and test environment follow in later slices.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `node scripts/test-bazel-overlay.mjs` fails on master with `bazel-cache-pilot must invoke Bazel` and passes on this slice
- [x] Same run no longer leaves a dummy-key `.bazelrc.user` behind

</details>

## Revert Plan

<details>
<summary>Revert Plan</summary>

- Safe to revert? Yes
- Revert command: `git revert <sha>`
- Post-revert steps: None
- Data migration? No

</details>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12672, opened 2026-09-15
[Bazel overlay](5) Wire BuildBuddy credentials into Bazel CI pilots

EdbertChan:
## Summary

Bazel does not expand `${BUILDBUDDY_API_KEY}` inside `.bazelrc`, so the pilots never authenticated to BuildBuddy even when the env var was set.

CI now writes a short-lived `.bazelrc.user` from the repo secret before the cache and RBE pilots run.

Auth headers were removed from the shared `build:remote` config so a literal placeholder key cannot break BES uploads.

## Review Claim

Bazel CI pilots authenticate to BuildBuddy via a GitHub secret written into ephemeral `.bazelrc.user`, without making Bazel a Mergify merge gate.

## Review Lane

policy

## Review Unit

tooling-policy

## Safety Invariant

The BuildBuddy API key is never committed; CI materializes it only into gitignored `.bazelrc.user` from `secrets.BUILDBUDDY_API_KEY`, and Mergify `merge_conditions` still do not require bazel pilots.

## Slice Rationale

Credential wiring is separated from generator/BUILD cache-key changes so each stack slice stays one review claim.

## Non-goals

Does not flip Mergify to require bazel-cache-pilot.
Does not enable RBE by default outside `--config=rbe`.
Does not change wrap-pnpm `local` tags, test sizes, or package BUILD generation.
Does not change product package behavior under pnpm.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `node scripts/test-bazel-overlay.mjs`
- [x] `node scripts/test-ci-workflow-merge-queue-policy.mjs`
- [x] `gh secret list` shows `BUILDBUDDY_API_KEY` on `Neko-Catpital-Labs/Invoker`
- [ ] After publish: `bazel-cache-pilot` streams to BuildBuddy without `Invalid API key`

</details>

## Revert Plan

<details>
<summary>Revert Plan</summary>

- Safe to revert? Yes
- Revert command: `git revert <sha>`
- Post-revert steps: Optional — rotate `BUILDBUDDY_API_KEY` if exposure is a concern
- Data migration? No

</details>

Depends-On: #12652

## Summary by CodeRabbit

* **Build & CI**
  * Added optional BuildBuddy remote caching and build-event services for supported CI jobs.
  * CI now configures authenticated remote services when credentials are available and continues without them when unavailable.
  * Standardized remote operation timeout to 10 minutes.
  * Moved credential handling out of shared build configuration into user- and CI-specific settings.

* **Validation**
[...]

PR #12409, opened 2026-09-14
[Submit latency hunt 8b] Bring 500-task plan intake under 200ms on the remote runner

EdbertChan:
## Summary

The remote runner still spent 765.88ms acknowledging a 500-task plan after the transaction optimization.

Plan intake now groups task and event persistence while retaining atomic failure behavior and the existing stored rows.

## Review Claim

Plan intake for a 500-task plan stays under the 200ms budget on the remote runner, with the same stored rows as before.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

The stored workflow, task, dependency, and event rows remain the same, and a failure partway through intake leaves no partial workflow. Repro runs use a throwaway owner, temp HOME, and temp database.

## Slice Rationale

This is one behavior slice for the remaining cost in a single plan-intake path. Proof-only work and handoff hygiene are excluded because they have no file changes to review.

## Non-goals

No plan YAML format change, no repro budget or assertion edits, no deferred persistence after acknowledgment, and no unrelated refactor.

### Before

```mermaid
graph TD
    A["Plan intake"] --> B["One persistence call per task and event"]
    B --> C["Per-item delta publication"]
```

### After

```mermaid
graph TD
    A["Plan intake"] --> B["Grouped task and event persistence"]
    B --> C["Atomic commit"]
    C --> D["Delta publication after commit"]
```

## Test Plan

<details>
<summary>Test Plan</summary>

- [ ] `pnpm --filter @invoker/app test --run src/__tests__/submit-latency-huge-plan.repro.test.ts src/__tests__/plan-submission-loader.test.ts`
- [ ] `pnpm --filter @invoker/workflow-core test`
- [ ] `pnpm --filter @invoker/data-store test`
- [ ] `pnpm run check:comments`

</details>

## Revert Plan

<details>
<summary>Revert Plan</summary>

- Safe to revert? Yes.
- Revert command: `git revert <published-commit-sha>`.
- Post-revert steps: Rerun the focused intake and neighboring tests.
- Data migration? No.

</details>

---

> [!NOTE]
> **Medium Risk**
> The custom YAML fast path can diverge from full YAML for non-conforming input (fallback applies), and the intake atomicity test documents behavior that depends on orchestrator/persistence transaction semantics.
> 
> **Overview**
[...]

PR #12862, opened 2026-09-17
[Fix master test drift](1) plan-parser freshness tests get a generic object error instead of the specific field error

EdbertChan:
## Summary

Plans that write task freshness as an inline map, like `freshness: { unknown: true }`, now get the real field error again.

A fast parsing shortcut added for large plans (#12409) stored inline maps and lists as plain text. The freshness check then only said "must be an object".

The shortcut now hands any value starting with `{` or `[` to the full YAML parser. Four plan-parser tests on master pass again.

## Review Claim

The plan fast path gives up on inline YAML maps and lists, so the full YAML parser reads them and the freshness check reports the specific field error.

## Review Lane

behavior

## Review Unit

routing

## Safety Invariant

Each fix only touches the files behind its own failing test group; that test file fails before and passes after, and every other test in the same package keeps its pre-change pass or fail result.

## Slice Rationale

One failure cause per workflow so each PR carries one reviewable claim; the other causes are separate workflows. This slice is one six-line change in `packages/app/src/plan-parser.ts`.

## Non-goals

- No change to the plan YAML format; #12409 also lists that as a non-goal.
- No change to the freshness check from #11534 or to the test file.
- Unrelated test files, refactors, and other failure causes listed in sibling workflows.
- Does not touch `quality / Bazel Overlay`; see "Inherited Failing Check" below.

## Test Plan

<details>
<summary>Test Plan</summary>

- [x] `cd packages/app && pnpm test -- src/__tests__/plan-parser.test.ts` with master's `plan-parser.ts`: `Tests  4 failed | 83 passed (87)` (the four `rejects invalid task freshness` cases)
- [x] `cd packages/app && pnpm test -- src/__tests__/plan-parser.test.ts` with this change: `Tests  87 passed (87)`

</details>

## Revert Plan

<details>

- Safe to revert? Yes
- Post-revert steps: None; the four freshness tests in `plan-parser.test.ts` fail again.
- Data migration? No

</details>

## Inherited Failing Check

`quality / Bazel Overlay` is red on this PR but is **not caused by this branch**. It fails identically on a clean `origin/master` checkout at `6b3a2bc4` with `Error: bazel-cache-pilot must invoke Bazel` (exit 1), and was already red at `7057570d`, `a6294ee1`, and `e65427ab`. This branch changes no input to that check — `git diff origin/master...HEAD -- scripts/test-bazel-overlay.mjs .github/workflows/ci.yml scripts/bazel/ .bazelrc .mergify.yml` is empty; the whole diff is six lines in `packages/app/src/plan-parser.ts`.

Root cause on master: `scripts/test-bazel-overlay.mjs` picks the pilot's Bazel step with `job.steps.find((step) => step.env?.BUILDBUDDY_API_KEY)`, which now matches the earlier `Configure BuildBuddy credentials` step added in #12672. That step only writes `.bazelrc.user` and never invokes `bazelisk`, so the probe's capture file is never written.

The fix is already authored and in flight as **#12844** ("[Bazel pilots](1) Point the overlay self-check at the step that runs Bazel"), the bottom of the open #12844–#12847 stack. Applying #12844's one-line change on top of this branch turns the check green locally (all four `PASS:` lines, exit 0), so it is proven sufficient. Vendoring it here would duplicate in-flight work and would add a second review unit (`scripts/**` is `tooling-policy`, this PR is `routing`), so it is deliberately left alone.


EdbertChan:

12862. #12862 — [Fix master test drift](1) plan-parser freshness tests get a generic object error instead of the specific field error ← this PR
   Base: master
   Link: https://github.com/Neko-Catpital-Labs/Invoker/pull/12862

